PR #1, opened 2016-03-01
Alison Cook

acook802:
Feedback assignment HW1. I could definitely use help with knowing where to provide more analysis, and how to clean graphs up for presentation.

Thanks!
Ali

### DIFF
--- a/HW-1.Rmd
+++ b/HW-1.Rmd
@@ -1,31 +1,31 @@
 ---
 title: "MATH 216 Homework 1"
-author: "WRITE YOUR NAME HERE"
+author: "Alison Cook"
 output: html_document
 ---
 
-```{r, echo=FALSE, message=FALSE, cache=TRUE}
+```{r, echo=FALSE, message=FALSE}
 # DO NOT EDIT THIS SECTION
-
 # Load packages
 suppressPackageStartupMessages(library(ggplot2))
 suppressPackageStartupMessages(library(dplyr))
 suppressPackageStartupMessages(library(lubridate))
-
+```
+```{r, echo=FALSE, message=FALSE, cache=TRUE}
 # Load data sets. Note this assumes this .Rmd files is in the same directory as
 # the .csv files.
 flights <- read.csv("data/flights.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df() %>%
-  mutate(date=as.Date(date))
+ tbl_df() %>%
+ mutate(date=as.Date(date))
 weather <- read.csv("data/weather.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df() %>%
-  mutate(date=as.Date(date))
+ tbl_df() %>%
+ mutate(date=as.Date(date))
 planes <- read.csv("data/planes.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df()
+ tbl_df()
 airports <- read.csv("data/airports.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df()
+ tbl_df()
 states <- read.csv("data/states.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df()
+ tbl_df()
 ```
 
 
@@ -36,10 +36,10 @@ states <- read.csv("data/states.csv", stringsAsFactors = FALSE) %>%
 
 Please indicate
 
-* Who you collaborated with:
-* Roughly how much time you spent on this HW:
-* What gave you the most trouble:
-* Any comments you have:
+* Who you collaborated with: Jacob Dixon and Andrew Holtz
+* Roughly how much time you spent on this HW: ~5 hours
+* What gave you the most trouble: Figuring how to represent data and how much analysis to perform on each question.
+* Any comments you have:Clarification in the questions would be helpful--I know they are intentionally open-ended, but some more ideas as to what you are looking for would be helpful. Also, I got caught up on really dumb small errors like a misplaced piping symbol or something. 
 
 
 
@@ -53,6 +53,35 @@ Please indicate
 * Which seasons did we tend to see the most and least delays of > 30 minutes.
 
 ```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
+# first find the total number of flights per day
+flights_per_day <- flights %>% 
+  group_by(date) %>% 
+  tally() %>% 
+  dplyr::rename(flights_per_day = n)
+flights_per_day
+
+# find number of flights delayed by >30 min on each day
+delayed_flights_per_day <- flights %>% 
+  group_by(date) %>% 
+  filter(dep_delay >30) %>% 
+  tally() %>% 
+  dplyr::rename(delayed_flights_per_day = n)
+delayed_flights_per_day
+
+# join data sets and compute the proportion of delayed flights per day
+prop_delayed_flights <- inner_join(flights_per_day, delayed_flights_per_day, by="date") %>% 
+                        mutate(prop = round(delayed_flights_per_day/flights_per_day,3))
+prop_delayed_flights
+
+
+#plot proportion of delayed flights
+ggplot(data=prop_delayed_flights, aes(x = date, y = prop)) +
+  geom_point() +
+  ggtitle("Delayed Flights") +
+  xlab("Date") +
+  ylab("Proportion of Flights Delayed")
+
+# separate by seasons and mark on plot (with color??)
 
 ```
 
@@ -66,6 +95,21 @@ Some people prefer flying on older planes.  Even though they aren't as nice,
 they tend to have more room.  Which airlines should these people favor?
 
 ```{r, echo=FALSE, fig.width=12, fig.height=6}
+# join planes with flights by "plane" to match plane to airline
+# left join to preserve all flights
+flights_and_planes <- left_join(flights, planes, by = "plane")
+flights_and_planes
+
+# plot histograms for each airline
+ggplot(data=flights_and_planes, aes(year)) +
+  geom_histogram() +
+  facet_wrap(~carrier, nrow=3)
+# this tells me the information, but is hard to interpret
+# boxplots?
+ggplot(data=flights_and_planes, aes(x = carrier, y = year)) +
+  geom_boxplot() +
+  labs(title = "Plane Age by Airline Carrier", x = "Carrier", y = "Year of Plane Model")
+# clean this up a lot
 
 ```
 
@@ -82,6 +126,58 @@ they tend to have more room.  Which airlines should these people favor?
 
 
 ```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
+# join flights and airports to determine states
+flights_and_airports <- left_join(flights, airports, by = c("dest" = "iata"))
+flights_and_airports
+
+# look at only southwest (code WN) by state
+southwest_dest <- flights_and_airports %>% 
+  filter(carrier == "WN") %>% 
+  group_by(state) %>% 
+  tally() %>% 
+  dplyr::rename(count_dest = n)
+southwest_dest
+
+# barplot representation
+# make this prettier
+ggplot(data=southwest_dest, aes(x = state, y = count_dest)) +
+  geom_bar(stat = "identity") +
+  labs(title = "Southwest Flights by State", x = "State", y = "Number of Flights")
+
+#this would be WAY cooler as a heatscale map of the US
+# why isn't it working??
+# states_map <- map_data("state")
+# ggplot(data = southwest_dest, aes(fill = as.factor(count_dest))) +
+#   geom_map(aes(map_id = state), map = states_map) +
+#   expand_limits(x = states_map$long, y = states_map$lat)
+
+#now look at unique flights
+#select by flight number
+southwest_dest_unique <- flights_and_airports %>% 
+  distinct(flight) %>% 
+  filter(carrier == "WN") %>% 
+  group_by(state) %>% 
+  tally() %>% 
+  dplyr::rename(count_dest = n)
+southwest_dest_unique
+
+ggplot(data=southwest_dest_unique, aes(x = state, y = count_dest)) +
+  geom_bar(stat = "identity") +
+  labs(title = "Southwest Flights by State", x = "State", y = "Number of Flights")
+
+#select by flight destination
+#why don't all of the states show up here?
+southwest_dest_unique_dest <- flights_and_airports %>%
+  distinct(dest) %>% 
+  filter(carrier == "WN") %>% 
+  group_by(state) %>% 
+  tally() %>% 
+  dplyr::rename(count_dest = n)
+southwest_dest_unique_dest
+
+ggplot(data=southwest_dest_unique_dest, aes(x = state, y = count_dest)) +
+  geom_bar(stat = "identity")+
+  labs(title = "Southwest Flights by State", x = "State", y = "Number of Flights")
 
 ```
 
@@ -94,7 +190,21 @@ they tend to have more room.  Which airlines should these people favor?
 What weather patterns are associated with the biggest departure delays?
 
 ```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
+# look at conditions for now?
+# join by both date and hour
+flights_and_weather <- left_join(flights, weather, by = c("date", "hour"))
+flights_and_weather
 
+#group flights by weather condition
+flights_and_weather <- flights_and_weather %>% group_by(conditions)
+
+#plot boxplot
+ggplot(data = flights_and_weather, aes(x = conditions, y = dep_delay)) +
+  geom_boxplot() +
+  labs(title = "Delays in Different Weather Conditions", x = "Conditions", y = "Length of Delay")
+
+# It seems like there is a better way to show this
+# Also, incorporate wind speed, etc
 ```
 
 
@@ -108,5 +218,31 @@ carrier flies to from Houston in the month of July.  Consider the `month()`
 function from the `lubridate` package.
 
 ```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
+flight_region <- left_join(flights_and_airports, states, by = "state") %>% 
+  mutate(month = month(date))
+
+flight_region
+
+
+#find flights in july total
+flight_region %>% 
+  filter(month == 7) %>% 
+  tally() 
+
+# 20,548 flights in july out of houston
+# group by region and tally
+flights_per_region <- flight_region %>% 
+  filter(month == 7) %>% 
+  group_by(region) %>% 
+  tally() %>% 
+  dplyr::rename(num_flights = n)
+
+flights_per_region <- mutate(flights_per_region, prop = round(num_flights/20548, 3))
+
+flights_per_region
+
+ggplot(data = flights_per_region, aes(x = region, y = prop)) +
+  geom_bar(stat = "identity") +
+  labs(title = "Regional Flights in July", x = "Region", y = "Proportion of Flights")
 
 ```


rudeboybert:
Sounds good. I'll be going over this on Thursday; expect feedback then.

HW-1.Rmd
```diff
@@ -36,10 +36,10 @@ states <- read.csv("data/states.csv", stringsAsFactors = FALSE) %>%
...
 Please indicate
 
-* Who you collaborated with:
-* Roughly how much time you spent on this HW:
-* What gave you the most trouble:
-* Any comments you have:
+* Who you collaborated with: Jacob Dixon and Andrew Holtz
+* Roughly how much time you spent on this HW: ~5 hours
+* What gave you the most trouble: Figuring how to represent data and how much analysis to perform on each question.
+* Any comments you have:Clarification in the questions would be helpful--I know they are intentionally open-ended, but some more ideas as to what you are looking for would be helpful. Also, I got caught up on really dumb small errors like a misplaced piping symbol or something.
```
rudeboybert: In terms of the analysis, think of it this way: you are a mother robin, your reader the baby robin. You are taking a raw worm (raw data) and providing something that’s easier to digest (interpretation). So imagine in this case the baby robin is someone in the airline industry. They want to know things, but they aren’t even sure what they want to know (hence the vague questions).  So the best advice I can offer is, you need to give these people the executive summary (through visualizations/tables/analysis and your written interpretation of the results) so that they can make decisions. 

What’s the right way to provide interpretation? There is none! Hence data analysis being part art as well as part science.  So there is only so much guidance I can offer. However, these vague guidelines is why I've set up the GitHub feedback loop in this class, so you can develop a sense of what's good and what's not as good interpretation.

Should the analysis be 100 words, or 10000 words? Again, no clear answer. Sometimes a clean crisp 100 word interpretation says more than a 10000 word one.  But one thing is for sure: less is more. All other things being equal, people will want to read the shorter report.

HW-1.Rmd
```diff
@@ -53,6 +53,35 @@ Please indicate
...
+
+
+#plot proportion of delayed flights
+ggplot(data=prop_delayed_flights, aes(x = date, y = prop)) +
+  geom_point() +
+  ggtitle("Delayed Flights") +
+  xlab("Date") +
+  ylab("Proportion of Flights Delayed")
+
+# separate by seasons and mark on plot (with color??)
```
rudeboybert: The plot in Question 1 looks good to start. However, your axes labels are a bit misinformative. The points don't reflect ALL delayed flights, but those delayed >= 30 minutes.

Also, I see you haven’t addressed the seasons issue yet (perhaps you haven’t gotten around to it yet).  One way is to explicitly associate dates to seasons.

Another way is to pick out “the signal from the noise” of your points. There definitely seems to be a trend, but its hard to pinpoint b/c of the variability of the points.

HW-1.Rmd
```diff
@@ -66,6 +95,21 @@ Some people prefer flying on older planes.  Even though they aren't as nice,
...
+# join planes with flights by "plane" to match plane to airline
+# left join to preserve all flights
+flights_and_planes <- left_join(flights, planes, by = "plane")
+flights_and_planes
+
+# plot histograms for each airline
+ggplot(data=flights_and_planes, aes(year)) +
+  geom_histogram() +
+  facet_wrap(~carrier, nrow=3)
+# this tells me the information, but is hard to interpret
```
rudeboybert: Good job faceting by airlines, but your histograms are hard to compare b/c they reflect counts. i.e. the biggest airlines: Continental, expressjet dominate b/c they have the most flights. If you make "density" histograms so that the histograms are all normalized to have sum of the areas of the boxes = 1, you'll be able to compare them better.

HW-1.Rmd
```diff
@@ -66,6 +95,21 @@ Some people prefer flying on older planes.  Even though they aren't as nice,
...
+# boxplots?
+ggplot(data=flights_and_planes, aes(x = carrier, y = year)) +
+  geom_boxplot() +
+  labs(title = "Plane Age by Airline Carrier", x = "Carrier", y = "Year of Plane Model")
+# clean this up a lot
```
rudeboybert: Good job! Both with this and the histogram, you get not only a notion of center of the distribution (i.e. mean/median age of planes), but also a notion of ... (what do you think)?

Now tell me, if I'm a flyer: all other things being equal, which airlines should I favor?

HW-1.Rmd
```diff
@@ -82,6 +126,58 @@ they tend to have more room.  Which airlines should these people favor?
...
+  geom_bar(stat = "identity") +
+  labs(title = "Southwest Flights by State", x = "State", y = "Number of Flights")
+
+#this would be WAY cooler as a heatscale map of the US
+# why isn't it working??
+# states_map <- map_data("state")
+# ggplot(data = southwest_dest, aes(fill = as.factor(count_dest))) +
+#   geom_map(aes(map_id = state), map = states_map) +
+#   expand_limits(x = states_map$long, y = states_map$lat)
+
```
rudeboybert: indeed it would be WAY cooler. We'll see this when we do maps. However, at first glance I think it may be b/c the state names in `states_map` don't match how they are named in `southwest_data`

HW-1.Rmd
```diff
@@ -82,6 +126,58 @@ they tend to have more room.  Which airlines should these people favor?
...
+  distinct(dest) %>% 
+  filter(carrier == "WN") %>% 
+  group_by(state) %>% 
+  tally() %>% 
+  dplyr::rename(count_dest = n)
+southwest_dest_unique_dest
+
+ggplot(data=southwest_dest_unique_dest, aes(x = state, y = count_dest)) +
+  geom_bar(stat = "identity")+
+  labs(title = "Southwest Flights by State", x = "State", y = "Number of Flights")
```
rudeboybert: Note your two plots have the same title, not allowing me to distinguish which is which.

HW-1.Rmd
```diff
@@ -82,6 +126,58 @@ they tend to have more room.  Which airlines should these people favor?
...
+southwest_dest_unique
+
+ggplot(data=southwest_dest_unique, aes(x = state, y = count_dest)) +
+  geom_bar(stat = "identity") +
+  labs(title = "Southwest Flights by State", x = "State", y = "Number of Flights")
+
+#select by flight destination
+#why don't all of the states show up here?
+southwest_dest_unique_dest <- flights_and_airports %>%
+  distinct(dest) %>%
```
rudeboybert: I'm not sure, but at first glance I think taking the distinct `dest` BEFORE you filter out the non-SW flights might have an unintended effect. i.e. you want to limit out SW flights first. I may be wrong however.

HW-1.Rmd
```diff
@@ -94,7 +190,21 @@ they tend to have more room.  Which airlines should these people favor?
...
+flights_and_weather <- left_join(flights, weather, by = c("date", "hour"))
+flights_and_weather
+
+#group flights by weather condition
+flights_and_weather <- flights_and_weather %>% group_by(conditions)
+
+#plot boxplot
+ggplot(data = flights_and_weather, aes(x = conditions, y = dep_delay)) +
+  geom_boxplot() +
+  labs(title = "Delays in Different Weather Conditions", x = "Conditions", y = "Length of Delay")
```
rudeboybert: Again, great job thinking boxplots, instead of just means. i.e. they give a sense of the WHOLE distribution, not just the center of the distribution.

However, your x-axis lables aren't helpful. Also,I can't really see what's going on in the box portions of the boxplots b/c the outliers are squishing the y-axis scale!

HW-1.Rmd
```diff
@@ -108,5 +218,31 @@ carrier flies to from Houston in the month of July.  Consider the `month()`
...
+  tally() %>% 
+  dplyr::rename(num_flights = n)
+
+flights_per_region <- mutate(flights_per_region, prop = round(num_flights/20548, 3))
+
+flights_per_region
+
+ggplot(data = flights_per_region, aes(x = region, y = prop)) +
+  geom_bar(stat = "identity") +
+  labs(title = "Regional Flights in July", x = "Region", y = "Proportion of Flights")
```
rudeboybert: Not sure if you've only gotten this far b/c of time constraints, but your plot here doesn't distinguish between carriers, and hence doesn't answer the question!

rudeboybert:
Overall you're definitely on the right track. Some polish here and there, and then your interpretations, and you'll be done. I also REALLY appreciate the clean coding: use of spaces, line breaks, keeping lines less than 80 characters long.

Hint: The way dplyr spews out tables is nasty. Consider:

* the `kable()` function from the `knitr` package
* [DT](https://rstudio.github.io/DT/) package

HW-1.Rmd
```diff
@@ -36,10 +36,10 @@ states <- read.csv("data/states.csv", stringsAsFactors = FALSE) %>%
 
 Please indicate
 
-* Who you collaborated with:
-* Roughly how much time you spent on this HW:
-* What gave you the most trouble:
-* Any comments you have:
+* Who you collaborated with: Jacob Dixon and Andrew Holtz
+* Roughly how much time you spent on this HW: ~5 hours
```
rudeboybert: You should go over the comments I have for Jacob #2  and Andrew #3 as well if you have time, since you worked together.

HW-1.Rmd
```diff
@@ -36,10 +41,10 @@ states <- read.csv("data/states.csv", stringsAsFactors = FALSE) %>%
...
 Please indicate
 
-* Who you collaborated with:
-* Roughly how much time you spent on this HW:
-* What gave you the most trouble:
-* Any comments you have:
+* Who you collaborated with: Jacob Dixon and Andrew Holtz
+* Roughly how much time you spent on this HW: ~12 hours
+* What gave you the most trouble: Figuring how to represent data and how much analysis to perform on each question.
+* Any comments you have: This was a LOT of time and work dealing with small small things. I felt like the amount of time I had to put in was disproportionate to what I was getting out of it.
```
rudeboybert: I feel you. I feel these things follow the 80/20 rule: 80% of the work gets done with 20% of your time/effort. Its drilling down to work the remaining 20% where things get messy. Please don't hesitate to ask for help if you feel like you're being hit by diminishing returns.

HW-1.Rmd
```diff
@@ -53,9 +58,58 @@ Please indicate
...
+
+# table
+datatable(prop_delayed_flights  %>% group_by(season)  %>% summarise(mean_prop_delays = round(mean(prop), 3)))
 ```
 
+### Delays of greater than 30 minutes are most common in the Summer
+Flights during the Summer months (June, July, August) had overall higher proportions of flight delays of over 30 minutes when compared to other seasons. A plot of counts of daily flights delays over 30 minutes shows an aggregate increase in delays during the summer months, and the corresponding boxplot and table corroborate this trend. The boxplot shows the distribution of delayed flights around the mean for each season. According to this plot, Summer is the most likely time to experience a delay, while Autumn is the least likely time to experience a delay. This could be due to the fact that Summer in the South is associated with extreme weather events such as hurricanes, while the Autumn is relatively mild in comparison. Therefore, if you are hoping to avoid a delay of more than 30 minutes out of Houston, Autumn flights are the best time to travel, followed by Winter flights, Spring flights, and finally Summer flights.
+
```
rudeboybert: Great job! Right down to the labels and titles (which a lot of people neglect). Since I need to say SOMETHING, I'll say the following:

* How about adding a `geom_smooth` to the first plot, so that you can pick out the signal from the noise?
* Good job using the `DT` package for the table. However, for a simple table like this, I think using DT is a bit like using an elephant gun to kill a mouse. We don't really need the interactivity of being able to sort by row and such. Here, a simple `knitr::kable()` call to convert an R table to a markdown table would've sufficed.
* Good job making hypotheses explaining the phenomenon. One analysis we could do is of the joint distribution between season and weather conditions.

HW-1.Rmd
```diff
@@ -53,9 +58,58 @@ Please indicate
...
+winter <- c(12, 1, 2)
+spring <- c(3, 4, 5)
+summer <- c(6, 7, 8)
+autumn <- c(9, 10, 11)
+
+prop_delayed_flights <- prop_delayed_flights %>% 
+  mutate(season = ifelse(month %in% winter, 'winter', 
+                         ifelse(month %in% spring, 'spring', 
+                                ifelse(month %in% summer, 'summer', 
+                                       ifelse(month %in% autumn, 'autumn', NA)))))
```
rudeboybert: I would've done the same thing here. I wish there were some function to do nested `ifelse()` statements better.

HW-1.Rmd
```diff
@@ -66,11 +120,39 @@ Some people prefer flying on older planes.  Even though they aren't as nice,
...
+# density shows proportionality rather than counts
+ggplot(data=flights_and_planes, aes(year)) +
+  geom_histogram(aes(y = ..density..)) +
+  facet_wrap(~carrier, nrow=3) +
+  labs(title = "Distribution of Fleet Age by Carrier", x = "Year of Plane Model", y = "Proportion of Fleet") 
+
+#organized by boxplot
+ggplot(data=flights_and_planes, aes(x = carrier, y = year)) +
+  geom_boxplot() +
+  labs(title = "Plane Age by Airline Carrier", x = "Carrier", y = "Year of Plane Model")
```
rudeboybert: I'm definitely a bigger fan of the boxplot than the histogram here. You can make comparisons between carriers with a single horizontal line (b/c all the action of a boxplot is long a single dimension), whereas with histograms the action is along two dimensions, so it takes more cognitive bandwidth to process the info. Great job.

Reach for the sky: consider ordering the boxplots in order of median as follows:

```r
ordered_carrier <- flights_and_planes %>% 
  group_by(carrier) %>% 
  summarise(median=median(year, na.rm=TRUE)) %>% 
  arrange(desc(median)) %>% 
  .[["carrier"]]

#organized by boxplot
ggplot(data=flights_and_planes, aes(x = factor(carrier, levels=ordered_carrier), y = year)) +
  geom_boxplot() +
  labs(title = "Plane Age by Airline Carrier", x = "Carrier", y = "Year of Plane Model")
```

HW-1.Rmd
```diff
@@ -66,11 +120,39 @@ Some people prefer flying on older planes.  Even though they aren't as nice,
...
+
+#table: number of planes by carrier for each year
```
rudeboybert: I would put a subsection header, or at least some text indicating what these tables are.

Also IMO, there's too much info in the 1st datatable to digest, I would just stick with the 2nd. but that's me.

HW-1.Rmd
```diff
@@ -66,11 +120,39 @@ Some people prefer flying on older planes.  Even though they aren't as nice,
...
... accounts for two of Envoy Air's planes. 224 Envoy planes did not have information about the model and year of plane associated with them, so it is possible that the vast majority of Envoy Air flights are on newer planes. According to the boxplot and accompanying data table, American Airlines has the most definitively older fleet, with many outliers made previous to 1980. 11402 planes were omitted in the boxplot and in calculating the mean age of fleets due to the fact that they did not have information about the model and year made associated with them. The boxplots give an indication of the range of age of each carrier's fleet, as well as the distribution. This could inform fliers of how likely they are to get an older fleet, given the size of the fleet and range of model years. Overall, if you are looking to fly on an older plane, American Airlines is the best airline to fly. The next best airline to fly is Delta, however, based on the data table that shows how many planes of each year each carrier has, Delta has been acquiring a few planes every year or so--which skews their fleet newer and increases your likelihood of flying on a newer plane as compared with American Airlines.
```
rudeboybert: Great job point out the potential biases induced by simply sweeping missing data under the rug. We'll talk more about this in class.

HW-1.Rmd
```diff
@@ -82,9 +164,80 @@ they tend to have more room.  Which airlines should these people favor?
...
+  dplyr::rename(count_dest = n)
+
+southwest_dest %>% arrange(desc(count_dest)) %>% datatable(.)
+
+#barplot ordered by frequency, not state
+southwest_dest$state <- factor(southwest_dest$state, 
+                               levels = southwest_dest$state[order(-southwest_dest$count_dest)])
+ggplot(data=southwest_dest, aes(x = state, y = count_dest)) +
+  geom_bar(stat = "identity") +
+  labs(title = "Southwest Flights by State", x = "State", y = "Number of Flights")
```
rudeboybert: Having both the table and the barplot is redundant. They provide essentially the same info; I would go with only one or the other. If you want to be able to get the exact values off the barchart, consider [interactive graphics](https://plot.ly/ggplot2/geom_bar/)

HW-1.Rmd
```diff
@@ -82,9 +164,80 @@ they tend to have more room.  Which airlines should these people favor?
...
+                 fill="#ffffff", color="#ffffff", size=0.15)
+g = g + geom_map(data=southwest_dest_states, map=us,
+                 aes(fill=count_dest, map_id=region),
+                 color="#ffffff", size=0.15)
+g = g + scale_fill_continuous(low='thistle2', high='darkblue', guide='colorbar') +  
+                 xlab("") + ylab("") + coord_map("albers", lat0 = 39, lat1 = 45)
+g = g + theme(panel.border = element_blank()) + theme(axis.ticks = element_blank()) + 
+                 theme(axis.text = element_blank()) +
+                 ggtitle("Southwest Flights out of Houston by State")
+g
```
rudeboybert: absolutely gorgeous. Way to go above and beyond the minimum requirements for this question. THIS is how you convey information.

HW-1.Rmd
```diff
@@ -82,9 +164,80 @@ they tend to have more room.  Which airlines should these people favor?
...
+southwest_dest_unique$state <- factor(southwest_dest_unique$state, 
+                levels = southwest_dest_unique$state[order(-southwest_dest_unique$count_dest)])
+ggplot(data=southwest_dest_unique, aes(x = state, y = count_dest)) +
+  geom_bar(stat = "identity") +
+  labs(title = "Unique Southwest Flights by State", x = "State", y = "Number of Flights")
+
+#select by flight destination
+southwest_dest_unique_dest <- flights_and_airports %>%
+  filter(carrier == "WN") %>% 
+  distinct(dest) %>%
```
rudeboybert: Ah! You fell into a trap here. Consider `dest==ABQ` or `dest==DAL`. There are more than 1 flights that fly these. Open the `flights` data set filtered to only SW flights. You'll see there are two ABQ flights. You had to do `distinct(flight)` here.

HW-1.Rmd
```diff
@@ -94,9 +247,33 @@ they tend to have more room.  Which airlines should these people favor?
 What weather patterns are associated with the biggest departure delays?
 
 ```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
-
+# look at categorical conditions 
+# join by both date and hour
+flights_and_weather <- left_join(flights, weather, by = c("date", "hour"))
+
+#group flights by weather condition
+flights_and_weather <- flights_and_weather %>% group_by(conditions)
```
rudeboybert: Minor nitpick: I'd try to bunch up operations on the same dataset as much as possible with piping, instead of having fragmented bits lying around. i.e. combine lines 252 and 255.

HW-1.Rmd
```diff
@@ -94,9 +247,33 @@ they tend to have more room.  Which airlines should these people favor?
...
+
+#plot boxplot
+ggplot(data = flights_and_weather, aes(x = conditions, y = dep_delay)) +
+  geom_boxplot() +
+  theme(axis.text.x = element_text(angle = 90, hjust = 1)) +
+  scale_y_log10() +
+  labs(title = "Delay Length by Weather Condition", x = "Condition", y = "Departure Delay (minutes)")
```
rudeboybert: Great job with the log10 transformation and the rotated axes-tick lables.  However, it would probably be a good idea to explicitly remove those rows with `delay <= 0` and comment this, b/c log10(x) doesn't exist for these values. Note, this also has implications for your conclusions. You are presenting average delay information *on those flights that did get delayed*, and not average delay information FOR ALL flights.

HW-1.Rmd
```diff
@@ -94,9 +247,33 @@ they tend to have more room.  Which airlines should these people favor?
...
+
```
rudeboybert: Ah ok, you make mention of the log for values less than 0 here. 

But as a flight manager, I might retort: Ok, so freezing rain has the biggest average delays. Does Houston get a lot of freezing rain? 

Good thing you're thinking about clear days. Indeed, a model to explain delays that incorporates just weather does not suffice, i.e. more variables definitely need to be included. Clear days tend to be in the summer, and summer is busy!

HW-1.Rmd
```diff
@@ -108,5 +285,24 @@ carrier flies to from Houston in the month of July.  Consider the `month()`
...
+  geom_bar(stat="identity", position="fill") +
+  labs(title = "Flights by Carrier in July of 2011", x = "Carrier", y = "Proportion of Flights") +
+  scale_fill_discrete(name = "Region", labels = c("Midwest", "Northeast", "South", "West"))
 
 ```
+
+### Regional flights vary by carrier in July
+20548 total flights left Houston on 15 different carriers in July, 2011. Some carriers served only one region--American Airlines (south), Alaska Airlines (West), JetBlue (Northeast), Frontier Airlines (West), Airtran Airways (South), and Mesa Airlines (South)--while others (Continental, Delta, ExpressJet, Envoy Air, SkyWest, United, US Airways, and Southwest) served airports nationwide. Gray areas on the plot indicate flights that did not have a state associated with the destination airport. Again, these flights can be attributed to Northwest Florida Beaches International Airport, and thus are still flights to a southern airport. Overall, this information can help fliers determine which carriers serve the regions they are traveling to, and inform them of airlines that are more likely to serve smaller airports or fly to specific regions more regularly. 
+
```
rudeboybert: Solid. The only thing I can think of is maybe including some info on the number of flights each bar consists of. So yeah, AA is 100% south, but is that 100% of 3 flights, or 100% of 1000 flights?

rudeboybert:
**Overall comments**: bang-up job, right down to the minor details. They may have seemed to be a royal PITA, but I, as the reader, got to reap the dividends.

Good coding style is large part. If I did have to come up with anything, its more nit-picky:

* try to merge operations on the same data set with piping
* create only the minimal number of new data frames necessary to get the job done.

My feedback is best viewed in [Split Screen Mode](https://github.com/Middlebury-Data-Science/HW-1/pull/1/files)

PR #2, opened 2016-03-01
Jacob Dixon 

dixonjacob:
This is the first draft of this homework assignment. It was a toughy.

### DIFF
--- a/HW-1.Rmd
+++ b/HW-1.Rmd
@@ -1,31 +1,31 @@
 ---
 title: "MATH 216 Homework 1"
-author: "WRITE YOUR NAME HERE"
+author: "Jacob Dixon"
 output: html_document
 ---
 
-```{r, echo=FALSE, message=FALSE, cache=TRUE}
+```{r, echo=FALSE, message=FALSE}
 # DO NOT EDIT THIS SECTION
-
 # Load packages
 suppressPackageStartupMessages(library(ggplot2))
 suppressPackageStartupMessages(library(dplyr))
 suppressPackageStartupMessages(library(lubridate))
-
+```
+```{r, echo=FALSE, message=FALSE, cache=TRUE}
 # Load data sets. Note this assumes this .Rmd files is in the same directory as
 # the .csv files.
 flights <- read.csv("data/flights.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df() %>%
-  mutate(date=as.Date(date))
+ tbl_df() %>%
+ mutate(date=as.Date(date))
 weather <- read.csv("data/weather.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df() %>%
-  mutate(date=as.Date(date))
+ tbl_df() %>%
+ mutate(date=as.Date(date))
 planes <- read.csv("data/planes.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df()
+ tbl_df()
 airports <- read.csv("data/airports.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df()
+ tbl_df()
 states <- read.csv("data/states.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df()
+ tbl_df()
 ```
 
 
@@ -36,10 +36,10 @@ states <- read.csv("data/states.csv", stringsAsFactors = FALSE) %>%
 
 Please indicate
 
-* Who you collaborated with:
-* Roughly how much time you spent on this HW:
-* What gave you the most trouble:
-* Any comments you have:
+* Who you collaborated with: Alison Cook
+* Roughly how much time you spent on this HW: 7 hours or so
+* What gave you the most trouble: Dumb little technical things; forgetting to remove an extra piping symbol at the end of a function, capitalization, etc. 
+* Any comments you have: I was a little confused by some of the questions and had to spend too much time doing things that reall shouldn't. Not really anything you can do about that though...
 
 
 
@@ -54,9 +54,31 @@ Please indicate
 
 ```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
 
+#Determine the number of flights per day
+per_day <- flights %>% 
+  group_by(date) %>% 
+  tally() %>% 
+  rename(num_of_flights = n)
+
+#Determine the number of flights delayed per day
+delayed <- flights %>% 
+  group_by(date) %>% 
+  filter(dep_delay > 30) %>% 
+  tally() %>% 
+  rename(num_delayed = n)
+
+#Merge the two and then calculate the proportion delayed per day
+joined_delay_and_date <- left_join(per_day,delayed, by = "date") %>% 
+  mutate(prop = round(num_delayed/num_of_flights,3))
+
+#Graph dat stuff. 
+ggplot(data=joined_delay_and_date, aes(x = date, y = prop))+
+  geom_point() +
+  labs(title = "Will My Flight Be Delayed by 30 Minutes or More?", x = "Date", y = "Proportion of Delayed Flights")
+
+#other idea: add color for seasons? Hmm...
+
 ```
-
-
 
 
 
@@ -66,6 +88,17 @@ Some people prefer flying on older planes.  Even though they aren't as nice,
 they tend to have more room.  Which airlines should these people favor?
 
 ```{r, echo=FALSE, fig.width=12, fig.height=6}
+
+#Join by the planes and group by the carrier
+planes_joinedto_flights <- left_join(flights, planes, by = "plane") %>% 
+  group_by(carrier)
+
+#Plot the carrier versus the year the plane was made 
+ggplot(data=planes_joinedto_flights, aes(x = carrier, y = year))+
+  geom_boxplot() +
+  labs(title = "How do I find an old plane?", x = "Airline Carrier", y = "Year the aircraft was built")
+
+  
 
 ```
 
@@ -83,6 +116,38 @@ they tend to have more room.  Which airlines should these people favor?
 
 ```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
 
+#First, where did the listed flights go to:
+airports_joinedto_flights <- left_join(flights, airports, by = c("dest" = "iata")) %>% 
+  filter(carrier =="WN") %>% 
+  group_by(state) %>% 
+  distinct(flight) %>% 
+  tally() %>% 
+  rename(unique_flights_per_state  = n)
+  
+airports_joinedto_flights
+ 
+ggplot(data = airports_joinedto_flights, aes(x = state, y = unique_flights_per_state)) +
+  geom_bar(stat = "identity") +
+  labs(title = "Where does Southwest fly?", x = "Destination State", y = "Number of Flights")
+
+
+
+
+#Join by the airport code and filter out only Southwest (WN) flights
+airports_joinedto_flights <- left_join(flights, airports, by = c("dest" = "iata")) %>% 
+  filter(carrier=="WN") %>% 
+  group_by(state) %>% 
+  tally() %>% 
+  rename(flights_per_state = n)
+
+airports_joinedto_flights
+
+ggplot(data = airports_joinedto_flights, aes(x = state, y = flights_per_state)) +
+  geom_bar(stat = "identity") +
+  labs(title = "Where does Southwest really fly?", x = "Destination State", y = "Number of Flights")
+
+
+
 ```
 
 
@@ -94,6 +159,29 @@ they tend to have more room.  Which airlines should these people favor?
 What weather patterns are associated with the biggest departure delays?
 
 ```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
+
+#Join the flights with weather info 
+flights_with_weather <- left_join(flights, weather, by = c("date", "hour") ) %>% 
+  group_by(conditions)
+
+ggplot(data = flights_with_weather, aes(x = conditions, y = dep_delay)) +
+  geom_boxplot() + 
+  scale_x_discrete(labels = abbreviate) +
+  labs(title = "Weather Conditions and Departure Delays", x = "Weather Condition Abbreviation", y = "Delay Length")
+
+
+#This looks jumbled and tough to read, but it is the right idea. 
+
+
+#I want to make bins with delays of various ranges, but I don't know how to make that kind of statement
+#it would probably need some ifelses 
+
+# delay_frequencies <- flights_with_weather %>% 
+#   group_by(dep_delay) %>% 
+#   tally() %>% 
+#   #something creating bins
+#   rename(delay_range = n)
+
 
 ```
 
@@ -108,5 +196,48 @@ carrier flies to from Houston in the month of July.  Consider the `month()`
 function from the `lubridate` package.
 
 ```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
+
+#Do some joining to get us started
+airports_to_flights_first <- left_join(flights, airports, by = c("dest" = "iata"))
+
+#Complete the join so we can have all the info we want
+airports_to_flights_to_states <- left_join(airports_to_flights_first, states, by = "state") %>% 
+  mutate(month(date))
+
+#Filter out July
+airports_to_flights_to_states %>% 
+  filter(month(date) == 7) %>% 
+  tally()
+
+#This gives 20,548
+#Now to figure out how many in each region
+
+flights_per_region <- airports_to_flights_to_states %>% 
+  filter(month(date) == 7) %>% 
+  group_by(region) %>% 
+  tally() %>% 
+  rename(number_of_flights = n)
+
+#Now to do a quick calculation and add a column for the proportion
+
+flights_per_region <- flights_per_region %>% 
+  mutate(proportion = round(number_of_flights/20548,3))
+
+#Make this a bar plot! 
+
+ggplot(data = flights_per_region, aes(x = region, y = proportion)) +
+  geom_bar(stat = "identity") +
+  labs(title = "Flights in July", x = "Region", y = "Proportion")
+
+
+# #I think this would work, but I don't know how to get it to work 
+# flights_in_july <- airports_to_flights_to_states %>% 
+#   group_by(region) %>%
+#   filter(month(date)) %>% 
+#   #somehow select for just july
+#   tally() %>%
+#   #divide by the total for all regions to get proportion
+#   rename(flights_per_region = n)
+
 
 ```


rudeboybert:
I'll be going over these on Thursday, so expect feedback then.  And yeppppp, toughy indeed; you're working with REAL data, not rounded-edges/training-wheeled/sandboxed data!

HW-1.Rmd
```diff
@@ -54,9 +54,31 @@ Please indicate
...
+joined_delay_and_date <- left_join(per_day,delayed, by = "date") %>% 
+  mutate(prop = round(num_delayed/num_of_flights,3))
+
+#Graph dat stuff. 
+ggplot(data=joined_delay_and_date, aes(x = date, y = prop))+
+  geom_point() +
+  labs(title = "Will My Flight Be Delayed by 30 Minutes or More?", x = "Date", y = "Proportion of Delayed Flights")
+
+#other idea: add color for seasons? Hmm...
+
```
rudeboybert: That is one way to account for seasonal trends. But also think, is there a way to pick out “the signal from the noise” of your points i.e. "seasonal trends". There definitely seems to be a trend, but its hard to pinpoint b/c of the variability of the points.

HW-1.Rmd
```diff
@@ -36,10 +36,10 @@ states <- read.csv("data/states.csv", stringsAsFactors = FALSE) %>%
 
 Please indicate
 
-* Who you collaborated with:
-* Roughly how much time you spent on this HW:
-* What gave you the most trouble:
-* Any comments you have:
+* Who you collaborated with: Alison Cook
+* Roughly how much time you spent on this HW: 7 hours or so
```
rudeboybert: You should go over the comments I wrote for Alison if you have time as well, since you worked together: #1

HW-1.Rmd
```diff
@@ -67,6 +89,17 @@ they tend to have more room.  Which airlines should these people favor?
...
+  group_by(carrier)
+
+#Plot the carrier versus the year the plane was made 
+ggplot(data=planes_joinedto_flights, aes(x = carrier, y = year))+
+  geom_boxplot() +
+  labs(title = "How do I find an old plane?", x = "Airline Carrier", y = "Year the aircraft was built")
+
+  
+
 ```
```
rudeboybert: Looks good. See my comment on line 43 of Alison's pull request about interpretation. #1

HW-1.Rmd
```diff
@@ -83,6 +116,38 @@ they tend to have more room.  Which airlines should these people favor?
...
+
+airports_joinedto_flights
+
+ggplot(data = airports_joinedto_flights, aes(x = state, y = flights_per_state)) +
+  geom_bar(stat = "identity") +
+  labs(title = "Where does Southwest really fly?", x = "Destination State", y = "Number of Flights")
+
+
+
 ```
```
rudeboybert: Looking good. See my comment [here](https://github.com/Middlebury-Data-Science/HW-1/pull/1#issuecomment-191763128) about 

* printing tables in R Markdown
* reordering bars in bar plots. i.e. not going "ALABAMA FIRST!"

HW-1.Rmd
```diff
@@ -95,6 +160,29 @@ What weather patterns are associated with the biggest departure delays?
...
 ```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
 
+#Join the flights with weather info 
+flights_with_weather <- left_join(flights, weather, by = c("date", "hour") ) %>% 
+  group_by(conditions)
+
+ggplot(data = flights_with_weather, aes(x = conditions, y = dep_delay)) +
+  geom_boxplot() + 
+  scale_x_discrete(labels = abbreviate) +
+  labs(title = "Weather Conditions and Departure Delays", x = "Weather Condition Abbreviation", y = "Delay Length")
```
rudeboybert: I didn't know you could do this! Amazing!

However, if you're going to do this, it would be useful for someone trying to read this plot to have a legend.

HW-1.Rmd
```diff
@@ -95,6 +160,29 @@ What weather patterns are associated with the biggest departure delays?
...
+#I want to make bins with delays of various ranges, but I don't know how to make that kind of statement
+#it would probably need some ifelses 
+
+# delay_frequencies <- flights_with_weather %>% 
+#   group_by(dep_delay) %>% 
+#   tally() %>% 
+#   #something creating bins
+#   rename(delay_range = n)
+
+
```
rudeboybert: Not quite sure what you mean, but yes, at times you need to use a morass of nested 
```r
ifelse(TESTA, A, 
  ifelse(TESTB, B, 
    ifelse(TEST, C, D)))
```
statements, which is an unfortunate lacking feature of dplyr. i.e. there is no `cases` command.

HW-1.Rmd
```diff
@@ -109,4 +197,47 @@ function from the `lubridate` package.
...
+flights_per_region <- airports_to_flights_to_states %>% 
+  filter(month(date) == 7) %>% 
+  group_by(region) %>% 
+  tally() %>% 
+  rename(number_of_flights = n)
+
+#Now to do a quick calculation and add a column for the proportion
+
+flights_per_region <- flights_per_region %>% 
+  mutate(proportion = round(number_of_flights/20548,3))
```
rudeboybert: Good practice note: its generally not a good idea to hard code a specific number, like 20548 here. If the dataset were to change, then this value would not update automagically. It's better to write code that computes this and then assigns to a variable.

HW-1.Rmd
```diff
@@ -109,4 +197,47 @@ function from the `lubridate` package.
...
+
+ggplot(data = flights_per_region, aes(x = region, y = proportion)) +
+  geom_bar(stat = "identity") +
+  labs(title = "Flights in July", x = "Region", y = "Proportion")
+
+
+# #I think this would work, but I don't know how to get it to work 
+# flights_in_july <- airports_to_flights_to_states %>% 
+#   group_by(region) %>%
+#   filter(month(date)) %>%
```
rudeboybert: forgot to `== 7` here

HW-1.Rmd
```diff
@@ -109,4 +197,47 @@ function from the `lubridate` package.
...
+#Now to do a quick calculation and add a column for the proportion
+
+flights_per_region <- flights_per_region %>% 
+  mutate(proportion = round(number_of_flights/20548,3))
+
+#Make this a bar plot! 
+
+ggplot(data = flights_per_region, aes(x = region, y = proportion)) +
+  geom_bar(stat = "identity") +
+  labs(title = "Flights in July", x = "Region", y = "Proportion")
```
rudeboybert: Note, this plot reflects no notion of carrier, hence does not answer the question!

rudeboybert:
Overall looks good.  I appreciate the clean, well spaced, well line-breaked, commented code. I'm going to go over it's mechanics with a finer tooth comb when I actually grade it next week.

HW-1.Rmd
```diff
@@ -36,10 +37,10 @@ states <- read.csv("data/states.csv", stringsAsFactors = FALSE) %>%
...
 Please indicate
 
-* Who you collaborated with:
-* Roughly how much time you spent on this HW:
-* What gave you the most trouble:
-* Any comments you have:
+* Who you collaborated with: Alison Cook
+* Roughly how much time you spent on this HW: 15 hours+
+* What gave you the most trouble: Unintuitive ggplot features. Trying to add a legend to a figure with abbreviated titles is (to me) impossible.
+* Any comments you have: I was confused by the ambiguity of some of the questions and had to spend too much time doing things that really I shouldn't have focused on.
```
rudeboybert: * I agree `ggplot2` is tough at times. Whereas `dplyr` is pretty straightforward and there is almost no doubt that it is a better way to learn data manipulation than base R, opinions still vary ALOT on ggplot2: [pro](http://varianceexplained.org/r/why-I-use-ggplot2/) and [con](http://simplystatistics.org/2016/02/11/why-i-dont-use-ggplot2/).
* (Keep in mind: of course I'm going to say this and I hope I don't sound too patronizing while saying it): I can empathize that the ambiguity is annoying, but it is not bad in a pedagogical sense. people don't always know what they want, so it's important that as an analyst you keep in constant touch and ask many many questions. For example, on a forestry project I'm working on, even though my collaborator came to me looking for stats help, it ends up being me asking more questions!

HW-1.Rmd
```diff
@@ -54,10 +55,44 @@ Please indicate
...
+#Merge the two and then calculate the proportion delayed per day (as a percent)
+joined_delay_and_date <- left_join(per_day,delayed, by = "date") %>% 
+  mutate(prop = round(num_delayed/num_of_flights*100,3))
+
+#Add month column
+joined_delay_and_date <- joined_delay_and_date %>% 
+  mutate(month = month(date))
+
+#add season column
+joined_delay_and_date <- joined_delay_and_date %>%
```
rudeboybert: Minor nit-pick: its easier to read if operations on the same data set are in one continuous block, instead of fragmented. I would merge your operations on lines 76 and 80 into a single block using piping.

HW-1.Rmd
```diff
@@ -54,10 +55,44 @@ Please indicate
...
+#Add month column
+joined_delay_and_date <- joined_delay_and_date %>% 
+  mutate(month = month(date))
+
+#add season column
+joined_delay_and_date <- joined_delay_and_date %>% 
+  mutate(season = ifelse(month == 12|month == 1|month == 2, "Winter",
+                         ifelse(month == 3|month == 4|month == 5, "Spring",
+                                ifelse(month == 6|month == 7|month == 8, "Summer",
+                                       ifelse(month == 9|month == 10|month == 11, "Fall", NA)))))
```
rudeboybert: I wish there were a better way to handle such situations than nested `ifelse` statements, like a `case` operator. unfortunately, I don't know of one.

HW-1.Rmd
```diff
@@ -54,10 +55,44 @@ Please indicate
...
+
+#Graph dat stuff. 
+ggplot(data=joined_delay_and_date, aes(x = season, y = prop, fill = season)) +
+  geom_boxplot() +
+  theme_tufte() +
+  guides(fill=FALSE) +
+  labs(title = "Will My Flight Be Delayed by 30 Minutes or More?", 
+       x = "Season", y = "Percentage of Delayed Flights")
```
rudeboybert: looks good, although I would argue that the color aesthetic, while looking pretty, adds no new information to the plot. I see you used the tufte theme. Tufte has a basic principle called the ink/information ratio. minimize ink but maximize information. So in this case, you're only increasing the numerator by using color, but not the denominator.

HW-1.Rmd
```diff
@@ -54,10 +55,44 @@ Please indicate
...
+  filter(dep_delay > 30) %>% 
+  tally() %>% 
+  rename(num_delayed = n)
+
+#Merge the two and then calculate the proportion delayed per day (as a percent)
+joined_delay_and_date <- left_join(per_day,delayed, by = "date") %>% 
+  mutate(prop = round(num_delayed/num_of_flights*100,3))
+
+#Add month column
+joined_delay_and_date <- joined_delay_and_date %>%
```
rudeboybert: Minor nit-pick: its easier to read if operations on the same data set are in one continuous block, instead of fragmented. I would merge your operations on lines 72 and 76 into a single block using piping. i.e.

```r
joined_delay_and_date <- left_join(per_day,delayed, by = "date") %>% 
  mutate(
    prop = round(num_delayed/num_of_flights*100,3),
    month = month(date)
    )
```

HW-1.Rmd
```diff
@@ -54,10 +55,44 @@ Please indicate
...
+#Graph dat stuff. 
+ggplot(data=joined_delay_and_date, aes(x = season, y = prop, fill = season)) +
+  geom_boxplot() +
+  theme_tufte() +
+  guides(fill=FALSE) +
+  labs(title = "Will My Flight Be Delayed by 30 Minutes or More?", 
+       x = "Season", y = "Percentage of Delayed Flights")
 
+```
+It is clear by looking at the figure above that the majority of flights delayed by more than thirty minutes take place during the spring and summer. The fall and winter experience the fewest relative delays at this time. To understand why these delays take place, one must think about where these flights are taking place. The region of the U.S. these flights are going to and departing from (the south, as seen below) is more heavily affected by summer thunderstorms and other weather, rather than snow in the winter. Cold weather, especially in the form of freezing rain is influential, but less so than spring and summer weather conditions.
```
rudeboybert: However, how many hours of freezing rain are there in Houston, one might retort.

HW-1.Rmd
```diff
@@ -67,11 +102,20 @@ they tend to have more room.  Which airlines should these people favor?
...
+planes_joinedto_flights <- left_join(flights, planes, by = "plane") %>% 
+  group_by(carrier)
 
+#Plot the carrier versus the year the plane was made 
+ggplot(data=planes_joinedto_flights, aes(x = carrier, y = year)) +
+  geom_boxplot() +
+  theme_tufte() +
+  labs(title = "How do I find an old plane?", x = "Airline Carrier", 
+       y = "Year the aircraft was built")
+
```
rudeboybert: See my [comment](https://github.com/Middlebury-Data-Science/HW-1/pull/1#discussion_r55703333) on Alison's pull request on sorting the boxplots by median, just to make the comparisons easier to swallow.

HW-1.Rmd
```diff
@@ -67,11 +102,20 @@ they tend to have more room.  Which airlines should these people favor?
...
+#Plot the carrier versus the year the plane was made 
+ggplot(data=planes_joinedto_flights, aes(x = carrier, y = year)) +
+  geom_boxplot() +
+  theme_tufte() +
+  labs(title = "How do I find an old plane?", x = "Airline Carrier", 
+       y = "Year the aircraft was built")
+  
 
+```
+The figure above shows the distribution of manufacture years for planes on the various airlines. It can be seen that Envoy Air (MQ) has a fleet comprised of all “older” aircraft. This is not one of the large carriers, however, so destinations will be more limited. American Airlines (AA) would likely be the best choice for a customer looking for older aircraft. They are a large airline with many destinations and have a fleet comprised of aircraft mostly built around 1990. Delta is very similar in its older aircraft age to American Airlines, but there are planes in Delta’s fleet that are much newer. All other airlines tend to have newer aircraft (post 1995). It should also be kept in mind that airline companies can retro-fit the interiors of their aircraft, potentially eliminating the benefit of being an older aircraft.
```
rudeboybert: But how many planes does MQ operate? two "older" aircraft, or two hundred?

HW-1.Rmd
```diff
@@ -83,8 +127,48 @@ they tend to have more room.  Which airlines should these people favor?
...
+
+#Get these sorted! 
+airports_joinedto_flights$state <- factor(airports_joinedto_flights$state,
+                                          levels = airports_joinedto_flights$state[order(-airports_joinedto_flights$unique_flights_per_state)])
+#Plot! With a nice theme! 
+ggplot(data = airports_joinedto_flights, aes(x = state, y = unique_flights_per_state)) +
+  geom_bar(stat = "identity") +
+  theme_tufte() +
+  labs(title = "Where does Southwest fly?", x = "Destination State", y = "Number of Flights")
+
```
rudeboybert: I forget who, but someone came up with a good disambiguation: "SW Flight Patterns" vs "SW Flights"

HW-1.Rmd
```diff
@@ -83,8 +127,48 @@ they tend to have more room.  Which airlines should these people favor?
...
+       x = "Destination State", y = "Number of Flights")
+
+
+```
+Listed flights can be thought of as unique flights, and these are illustrated in the first figure above. These are flights with unique flight numbers. For example, if American Airlines flight 428 flies to Dallas, TX four times in one day, it will only be counted once for that day.  This is because it is the same flight, just at different times. The same is true for Southwest flights seen above. All the states flown to can be seen in the figures above, with Texas having the most flights. The number of unique flights in the top figure can be contrasted with the figure below that shows *all* of the flights that Southwest makes in one year. While a substantial number of unique flights go to Texas, an even larger difference can be seen in the total number of flights to Texas compared with the other states. It should be noted that “NA” is not a state, rather an airport in Florida that was not combined with other Florida flights because of its unlisted airport code. This addition does not change Florida’s ranking, nor the extreme disparity between Texas and any other state.
```
rudeboybert: Good job `distinct`ing by flight, and not dest. For example, there are more than one flight pattern that SW flies to `ABQ` albuquerque.

HW-1.Rmd
```diff
@@ -95,9 +179,24 @@ What weather patterns are associated with the biggest departure delays?
...
+#Graph that. Log transform to make it readable (removes delays =< 0)
+ggplot(data = flights_with_weather, aes(x = conditions, y = dep_delay)) +
+  geom_boxplot() + 
+  theme_tufte() +
+  theme(axis.text.x = element_text(angle = 90, vjust = 0.25, hjust = 1)) +
+  scale_y_log10() +
+  labs(title = "What does the weather mean for my flight?", 
+       x = "Weather Condition", y = "Delay Length (minutes)")
 
+#I could not get the legend to show up with abbreviated titles so I have rotated them instead.
```
rudeboybert: That's fine. At some point you got to give up on doing things perfectly and move on. My manager at Google kept telling us that "perfectionism = failure to launch" and this quote got bandied about Silicon Valley alot (to the point of nausea at times) [Done is better than perfect](http://www.fastcompany.com/3001533/truth-about-being-done-versus-being-perfect)

HW-1.Rmd
```diff
@@ -95,9 +179,24 @@ What weather patterns are associated with the biggest departure delays?
...
+  theme(axis.text.x = element_text(angle = 90, vjust = 0.25, hjust = 1)) +
+  scale_y_log10() +
+  labs(title = "What does the weather mean for my flight?", 
+       x = "Weather Condition", y = "Delay Length (minutes)")
 
+#I could not get the legend to show up with abbreviated titles so I have rotated them instead. 
+
+
+```
+The figure above shows the weather conditions that were recorded at the same time (within the hour) as flights were delayed. This is an association, rather than a causal relationship as all types of weather, even “clear” show a delay. It should also be noted that the delay length scale is logarithmic, moving up the axis means a dramatic increase in delay length for the traveler. The weather patterns to take note of are “freezing rain” and “light freezing rain” as these are most likely to cause the largest delay. That being said, these are likely to only occur during the winter, a time that has the fewest (proportionally) delays. The location of the destination and departure airport should be considered as well as the season and current weather conditions before predicting a possible delay.
```
rudeboybert: Good that you are not assert causation. There are DEFINITELY unaccounted for confounders in this model. 

Also, they occur only in the winter, but how often? Some notion of sample size would help.

As for departure airport, the dataset is only flights leaving Houston. Destination? I guess if a flight is coming in from further away, there is a higher chance of backlog...

HW-1.Rmd
```diff
@@ -95,9 +179,24 @@ What weather patterns are associated with the biggest departure delays?
...
+#Join the flights with weather info 
+flights_with_weather <- left_join(flights, weather, by = c("date", "hour") ) %>% 
+  group_by(conditions)
 
+#Graph that. Log transform to make it readable (removes delays =< 0)
+ggplot(data = flights_with_weather, aes(x = conditions, y = dep_delay)) +
+  geom_boxplot() + 
+  theme_tufte() +
+  theme(axis.text.x = element_text(angle = 90, vjust = 0.25, hjust = 1)) +
+  scale_y_log10() +
```
rudeboybert: See my note on Alison's [page](https://github.com/Middlebury-Data-Science/HW-1/pull/1#discussion_r55706268) about logging values less than 0.

rudeboybert:
**OVERALL COMMENTS**: Things look good. Good coding style; just try to

* try to merge operations on the same data set with piping
* create only the minimal number of new data frames necessary to get the job done.

My feedback is best viewed in [Split screen mode](https://github.com/Middlebury-Data-Science/HW-1/pull/2/files)

PR #3, opened 2016-03-02
Andrew Holtz

amholtz:
These are changes done as a first attempt at Homework assignment #1 for feedback.

### DIFF
--- a/HW-1.Rmd
+++ b/HW-1.Rmd
@@ -1,31 +1,33 @@
 ---
 title: "MATH 216 Homework 1"
-author: "WRITE YOUR NAME HERE"
+author: "Andrew Holtz"
 output: html_document
 ---
 
-```{r, echo=FALSE, message=FALSE, cache=TRUE}
+```{r, echo=FALSE, message=FALSE}
 # DO NOT EDIT THIS SECTION
 
 # Load packages
 suppressPackageStartupMessages(library(ggplot2))
 suppressPackageStartupMessages(library(dplyr))
 suppressPackageStartupMessages(library(lubridate))
+```
 
+```{r, echo=FALSE, message=FALSE, cache=TRUE}
 # Load data sets. Note this assumes this .Rmd files is in the same directory as
 # the .csv files.
 flights <- read.csv("data/flights.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df() %>%
-  mutate(date=as.Date(date))
+ tbl_df() %>%
+ mutate(date=as.Date(date))
 weather <- read.csv("data/weather.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df() %>%
-  mutate(date=as.Date(date))
+ tbl_df() %>%
+ mutate(date=as.Date(date))
 planes <- read.csv("data/planes.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df()
+ tbl_df()
 airports <- read.csv("data/airports.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df()
+ tbl_df()
 states <- read.csv("data/states.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df()
+ tbl_df()
 ```
 
 
@@ -36,10 +38,10 @@ states <- read.csv("data/states.csv", stringsAsFactors = FALSE) %>%
 
 Please indicate
 
-* Who you collaborated with:
-* Roughly how much time you spent on this HW:
-* What gave you the most trouble:
-* Any comments you have:
+* Who you collaborated with: Mohamed Hussein
+* Roughly how much time you spent on this HW: 9 hours
+* What gave you the most trouble: Problem number 4- I am not sure if I did this correctly. I did have a lot of trouble joining two tables for this problem. I had to join delaytime and weatherpattern by two variables "date" and "hour". I learned about full_join and using the c functino to do this. 
+* Any comments you have: I was confused for four and five what kind of conclusion I was actually suppose to make a figure for. It might be because I do not have great experience making figures yet.
 
 
 
@@ -54,6 +56,24 @@ Please indicate
 
 ```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
 
+total_flights <- select(flights, date, dep_delay) %>% 
+  group_by(date) %>% 
+  tally() %>% 
+  rename(num_flights = n)
+View(total_flights)
+  
+delay_flights <- select(flights, date, dep_delay) %>%
+  filter(dep_delay > 30) %>% 
+  group_by(date) %>% 
+  tally() %>% 
+  rename(num_delayed_flights = n)
+
+prop_delayed<- left_join(total_flights,delay_flights, by = "date") %>% 
+  mutate(prop = round(num_delayed_flights / num_flights , 3))
+
+ggplot(data=prop_delayed, aes(x = date, y = prop)) + ggtitle("Proportion of Flights delayed by more than 30 minutes on each day for one year") + 
+  ylab("Proportion") + xlab("Date") + geom_line()
+
 ```
 
 
@@ -67,12 +87,33 @@ they tend to have more room.  Which airlines should these people favor?
 
 ```{r, echo=FALSE, fig.width=12, fig.height=6}
 
+carrier_plane <- select(flights,carrier,plane)
+plane_year <- select(planes, plane, year)
+carrier_year <- left_join(carrier_plane, plane_year, by= "plane")
+
+total_planes <- carrier_year %>% 
+  filter(!is.na(year)) %>%
+  group_by(carrier) %>% 
+  tally() %>% 
+  rename(total_num_planes = n)
+
+old_planes <- carrier_year %>%
+  filter(!is.na(year)) %>%
+  group_by(carrier) %>% 
+  filter(year<1980) %>% 
+  tally() %>% 
+  rename(num_old_planes = n)
+View(old_planes)
+
+prop_old_plane <- left_join(old_planes, total_planes, by= "carrier") %>% 
+  mutate(num_new_planes = total_num_planes - num_old_planes) %>% 
+  mutate(prop = round(num_old_planes / total_num_planes , 3)) 
+ 
+
+ggplot(data = prop_old_plane, aes(x= carrier, y= prop)) + ggtitle("Proportion of planes that are older than 1980 per airline") +
+  xlab("Airline") + ylab("Proportion") + geom_bar(stat="identity")
+
 ```
-
-
-
-
-
 
 
 ## Question 3:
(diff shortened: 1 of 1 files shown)


rudeboybert:
Got it thanks. I'll be going over these on Thursday.

rudeboybert:
Heads up. Removing the `View()` functions calls solved it. I can compile your `.Rmd` file, the output of which I'll be going over tomorrow.

amholtz:
Great to hear! Thanks.

HW-1.Rmd
```diff
@@ -54,6 +56,24 @@ Please indicate
...
+  group_by(date) %>% 
+  tally() %>% 
+  rename(num_delayed_flights = n)
+
+prop_delayed<- left_join(total_flights,delay_flights, by = "date") %>% 
+  mutate(prop = round(num_delayed_flights / num_flights , 3))
+
+ggplot(data=prop_delayed, aes(x = date, y = prop)) + ggtitle("Proportion of Flights delayed by more than 30 minutes on each day for one year") + 
+  ylab("Proportion") + xlab("Date") + geom_line()
+
```
rudeboybert: Plot is looking good, however you still haven't answered the seasons question. You can either associate seasons to each of the dates. But also think, is there a way to pick out “the signal from the noise” of your points i.e. "seasonal trends". There definitely seems to be a trend, but its hard to pinpoint b/c of the variability of the points.

HW-1.Rmd
```diff
@@ -67,12 +87,33 @@ they tend to have more room.  Which airlines should these people favor?
 
 ```{r, echo=FALSE, fig.width=12, fig.height=6}
 
-```
+carrier_plane <- select(flights,carrier,plane)
+plane_year <- select(planes, plane, year)
+carrier_year <- left_join(carrier_plane, plane_year, by= "plane")
```
rudeboybert: good job selecting the variables before the join. Joins work much faster with smaller datasets.

HW-1.Rmd
```diff
@@ -67,12 +87,33 @@ they tend to have more room.  Which airlines should these people favor?
...
 
+total_planes <- carrier_year %>% 
+  filter(!is.na(year)) %>%
```
rudeboybert: Whenever dropping missing data, one has to be careful. Maybe the data is missing b/c the planes are so old no one knows when they are from! One should always acknowledge in the report if ever values are dropped.

HW-1.Rmd
```diff
@@ -67,12 +87,33 @@ they tend to have more room.  Which airlines should these people favor?
...
+  tally() %>% 
+  rename(num_old_planes = n)
+View(old_planes)
 
+prop_old_plane <- left_join(old_planes, total_planes, by= "carrier") %>% 
+  mutate(num_new_planes = total_num_planes - num_old_planes) %>% 
+  mutate(prop = round(num_old_planes / total_num_planes , 3)) 
+ 
 
+ggplot(data = prop_old_plane, aes(x= carrier, y= prop)) + ggtitle("Proportion of planes that are older than 1980 per airline") +
```
rudeboybert: This plot does answer the question, but barely. What if an airline had all its planes from 1979, and the other from 1981? Then they'd show up as 100% and 0% in your bars.  i.e. you're only focusing on measures of center. When describing a distribution, people always try to convey some notion of center. What's the next notion people should try to convey?

HW-1.Rmd
```diff
@@ -83,6 +124,35 @@ they tend to have more room.  Which airlines should these people favor?
...
+state_sw <- select(flights, dest, carrier) %>% 
+  filter(carrier == "WN")
+airport_state <- select(airports, iata, state)
+
+state_carrier_airport <- left_join(state_sw, airport_state, by = c("dest" = "iata")) %>% 
+  group_by(state) %>% 
+  filter(!is.na(state)) %>% 
+  tally()
+
+ggplot(data = state_carrier_airport, aes(x= state,y = n, fill = state)) + xlab("State") + ylab("Frequency") + ggtitle("Southwest Airlines most flied States") + geom_bar(stat="identity")
```
rudeboybert: While people like looking a colors, ask yourself: what does the color provide to the plot that isn't already provided by the bars? I generally like to take a "less is more" approach to plots. i.e. use the minimal amount of ink/features to get the point across.

HW-1.Rmd
```diff
@@ -83,6 +124,35 @@ they tend to have more room.  Which airlines should these people favor?
...
+#This leaves you with the frequency of flights to each city for each day
+airport_state <- select(airports, iata, state)
+state_freq <- left_join(state_sw_date, airport_state, by = c("dest" = "iata")) %>% 
+  #combine frequency of airport per day with state
+  select(dest, state) %>% #remove n variable
+  filter(!is.na(state)) %>% 
+  group_by(state) %>% 
+  tally()
+
+ggplot(data = state_freq, aes(x= state, y = n)) + xlab("State") + ylab("Frequency") + ggtitle("Southwest Airlines most flied States per day") + geom_bar(stat="identity")
```
rudeboybert: Something seems fishy about this plot. AL, AR, AZ, CO, IL, etc. all have the exact same number?

HW-1.Rmd
```diff
@@ -95,6 +165,48 @@ What weather patterns are associated with the biggest departure delays?
...
+                           
+                                  
+plot_hd <- ggplot(data = weather_pattern_delay, aes(x= humidity, y = avg_delay)) + geom_point()
+plot_dd <- ggplot(data = weather_pattern_delay, aes(x= dew_point, y = avg_delay)) + geom_point()
+plot_gd <- ggplot(data = weather_pattern_delay, aes(x= gust_speed, y = avg_delay)) + geom_point()
+
+plot_hd + xlab("Humidity(%)") + ylab("Delay(min)") + ggtitle("Biggest Delays by Humidity")
+plot_dd + xlab("Dew Point") + ylab("Delay(min)") + ggtitle("Biggest Delays by Dew Point")
+plot_gd + xlab("Gust Speed(mph)") + ylab("Delay(min)") + ggtitle("Biggest Delays by Gust Speed")
+
```
rudeboybert: In your report, you should definitely indicate why you chose these three weather patterns.

HW-1.Rmd
```diff
@@ -95,6 +165,48 @@ What weather patterns are associated with the biggest departure delays?
...
+#To accomplish conditions
+delay_conditions <- select(weather_pattern_delay, conditions, avg_delay) %>%
+  filter(!is.na(avg_delay)) %>% 
+  filter(!is.na(conditions)) %>% 
+  ungroup() %>% 
+  arrange(conditions) %>% 
+  group_by(conditions) %>% 
+  summarise(avg_delay = mean(avg_delay))
+
+ggplot(data = delay_conditions, aes(x= conditions, y = avg_delay, fill = conditions)) + xlab("Conditions") + geom_bar(stat="identity", position="stack") + ylab("Delay(min)") + ggtitle("Biggest Delays by Weather Condition")
```
rudeboybert: * Same remark as above about color.
* assuming your coding is right (which I'll go over when I actually grade), again, this plot does answer the question i.e it reports avg delay. For a lot of people, averages are all they can handle. But let's try to reach a little higher. After the average, what notion of a set of values should I be trying to convey?
* Also, your x-axes labels aren't helpful.

HW-1.Rmd
```diff
@@ -109,4 +221,21 @@ function from the `lubridate` package.
...
+state_carrier_airport <- left_join(state_carrier_airport, states, by = "state") 
+
+region_freq <- count(state_carrier_airport, carrier, region) %>% 
+  rename(Frequency = n) %>% 
+  filter(!is.na(region))
+
+g <- ggplot(data = region_freq, aes(x= carrier, y = Frequency, fill = region))
+  
+g + geom_bar(stat="identity", position="fill") + coord_flip() + xlab("Airline") + ggtitle("Proportion of Flights to 4 regions by airline in the month of July")
+#Question: How can I rename the observations for region- i.e. "south" to "South".
```
rudeboybert: Great looking plot! Not that its the right answer, but this is what I would've done!

To answer your question: You need to rename the variable in the actual dataset using `dplyr`. I dunno of a way to do this in `ggplot`

HW-1.Rmd
```diff
@@ -36,10 +38,10 @@ states <- read.csv("data/states.csv", stringsAsFactors = FALSE) %>%
...
 Please indicate
 
-* Who you collaborated with:
-* Roughly how much time you spent on this HW:
-* What gave you the most trouble:
-* Any comments you have:
+* Who you collaborated with: Mohamed Hussein
+* Roughly how much time you spent on this HW: 9 hours
+* What gave you the most trouble: Problem number 4- I am not sure if I did this correctly. I did have a lot of trouble joining two tables for this problem. I had to join delaytime and weatherpattern by two variables "date" and "hour". I learned about full_join and using the c functino to do this. 
+* Any comments you have: I was confused for four and five what kind of conclusion I was actually suppose to make a figure for. It might be because I do not have great experience making figures yet.
```
rudeboybert: yeah, the questions are pretty vague, hence knowing what conclusion to reach is pretty vague. Here is what I told Alison:

In terms of the analysis, think of it this way: you are a mother robin, your reader the baby robin. You are taking a raw worm (raw data) and providing something that’s easier to digest (interpretation). So imagine in this case the baby robin is someone in the airline industry. They want to know things, but they aren’t even sure what they want to know (hence the vague questions). So the best advice I can offer is, you need to give these people the executive summary (through visualizations/tables/analysis and your written interpretation of the results) so that they can make decisions.

What’s the right way to provide interpretation? There is none! Hence data analysis being part art as well as part science. So there is only so much guidance I can offer. However, these vague guidelines is why I've set up the GitHub feedback loop in this class, so you can develop a sense of what's good and what's not as good interpretation.

Should the analysis be 100 words, or 10000 words? Again, no clear answer. Sometimes a clean crisp 100 word interpretation says more than a 10000 word one. But one thing is for sure: less is more. All other things being equal, people will want to read the shorter report.

HW-1.Rmd
```diff
@@ -36,10 +38,10 @@ states <- read.csv("data/states.csv", stringsAsFactors = FALSE) %>%
 
 Please indicate
 
-* Who you collaborated with:
-* Roughly how much time you spent on this HW:
-* What gave you the most trouble:
-* Any comments you have:
+* Who you collaborated with: Mohamed Hussein
+* Roughly how much time you spent on this HW: 9 hours
+* What gave you the most trouble: Problem number 4- I am not sure if I did this correctly. I did have a lot of trouble joining two tables for this problem. I had to join delaytime and weatherpattern by two variables "date" and "hour". I learned about full_join and using the c functino to do this.
```
rudeboybert: Not quite sure what you mean, but at first glance I think you did your joins right. ask me in person the next time I see you.

rudeboybert:
Overall things are shaping nicely. Please try to break lines so that they don't exceed 80 characters. You can have a 80 character marker line displayed if you go to Tools -> Global Options -> Code -> Display.

Also, reordering bar plots by frequency makes for nice aesthetics. Not doing this for state-level data is known as the "ALABAMA FIRST" phenomenon. Check out this [link](http://www.reed.edu/data-at-reed/resources/R/reordering_geom_bar.html).

HW-1.Rmd
amholtz: I can have a distribution similar to the RNAseq table, with 1980 as the center.

rudeboybert: RNAseq? Where was this from?

amholtz: Oh, sorry.  This was just for my own thought process, a thought I had that I didn't want to forget so I left a reminder for myself.

PR #4, opened 2016-03-02
Ryan Clement

rkclement:
Doing this for the sake of trying it out and keeping up with things. Feel free to not give comments if you need to devote time to students actually taking the class, though.

### DIFF
--- a/HW-1.Rmd
+++ b/HW-1.Rmd
@@ -151,7 +151,7 @@ as.data.frame(uniqDestinations)
 uniqDestinations$state <- factor(uniqDestinations$state, levels = uniqDestinations$state)
 
 g <- ggplot(filter(uniqDestinations, !is.na(state)), aes(x = state, y = n))
-g + geom_bar(stat="identity") + coord_flip() + labs(title = "States to which Listed Southwest Flights Tend to Fly")
+g + geom_bar(stat = "identity") + coord_flip() + labs(title = "States to which Listed Southwest Flights Tend to Fly")
 
 # Next join all flights table, not just listed ones, and graph.  
 allSW <- swFlights %>% 
@@ -170,11 +170,9 @@ as.data.frame(allDestinations)
 allDestinations$state <- factor(allDestinations$state, levels = allDestinations$state)
 
 g <- ggplot(filter(allDestinations, !is.na(state)), aes(x = state, y = n))
-g + geom_bar(stat="identity") + coord_flip() + labs(title = "States to which All Southwest Flights Tend to Fly")
+g + geom_bar(stat = "identity") + coord_flip() + labs(title = "States to which All Southwest Flights Tend to Fly")
 
 ```
-
-
 
 
 
--- a/HW-1.Rmd
+++ b/HW-1.Rmd
@@ -1,31 +1,31 @@
----
+ca---
 title: "MATH 216 Homework 1"
-author: "WRITE YOUR NAME HERE"
+author: "Ryan Clement"
 output: html_document
 ---
 
-```{r, echo=FALSE, message=FALSE, cache=TRUE}
+```{r, echo=FALSE, message=FALSE}
 # DO NOT EDIT THIS SECTION
-
 # Load packages
 suppressPackageStartupMessages(library(ggplot2))
 suppressPackageStartupMessages(library(dplyr))
 suppressPackageStartupMessages(library(lubridate))
-
+```
+```{r, echo=FALSE, message=FALSE, cache=TRUE}
 # Load data sets. Note this assumes this .Rmd files is in the same directory as
 # the .csv files.
 flights <- read.csv("data/flights.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df() %>%
-  mutate(date=as.Date(date))
+ tbl_df() %>%
+ mutate(date=as.Date(date))
 weather <- read.csv("data/weather.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df() %>%
-  mutate(date=as.Date(date))
+ tbl_df() %>%
+ mutate(date=as.Date(date))
 planes <- read.csv("data/planes.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df()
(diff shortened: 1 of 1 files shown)


rudeboybert:
I think I should be able to give it a look. So far only a fraction of the students requested feedback.

HW-1.Rmd
```diff
@@ -36,10 +36,14 @@ states <- read.csv("data/states.csv", stringsAsFactors = FALSE) %>%
...
-* What gave you the most trouble:
-* Any comments you have:
+* Who you collaborated with: 
+*no one*
+* Roughly how much time you spent on this HW: 
+*3 hours? It was spread out in little chunks over the whole week...*
+* What gave you the most trouble: 
+*Trying to answer Question 2 graphically and get the graph to a point where I could interpret it properly.* 
+* Any comments you have: 
+*Some things weren't clear and could have been included in a "codebook" (e.g. I didn't realize all the flights were taking off from Houston, so I spent a lot of time puzzling over how to filter for the flights from Houston in the last question.)*
```
rudeboybert: true. I'll make a note for next time.

HW-1.Rmd
```diff
@@ -52,54 +56,148 @@ Please indicate
...
+# Next we join these two in order to get the proportion per day.
+flightsDelay_perDay <- left_join(flightsDelay_perDay, flights_perDay, by = "date") %>% 
+  mutate(prop_delay = round(delayed_per_day/flights_per_day, 3))
+
+g <- ggplot(data = flightsDelay_perDay, aes(x = date, y = prop_delay))
+g + geom_point() + 
+  labs(title = "Proportion of Flights Delayed More Than 30 Minutes", 
+                        x     = "Date", 
+                        y     = "Proportion Delayed")
+```
```
rudeboybert: I'd add a `stat_smooth()` to this plot so that the seasonality trend is really apparent.

HW-1.Rmd
```diff
@@ -52,54 +56,148 @@ Please indicate
...
+
+# We join the table of planes flown by each carrier to the year of each plane from the planes table.
+fleet <- planes %>% 
+  select(plane, year) %>%
+  left_join(fleet, ., by = "plane")
+
+# We plot the distribution of each carrier's fleet by plane manufacture year. Note that we remove planes 
+# where year data is missing. We separate the histograms into separate facets to make differences clearer.
+
+suppressPackageStartupMessages(library(RColorBrewer))
```
rudeboybert: It's considered good practice to have all your package loadings in the same place (in the beginning) so that people can see in one shot which packages they need installed.

HW-1.Rmd
```diff
@@ -52,54 +56,148 @@ Please indicate
...
+  tally () %>%
+  arrange(n)
 
-```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
+as.data.frame(allDestinations)
+
+allDestinations$state <- factor(allDestinations$state, levels = allDestinations$state)
+
+g <- ggplot(filter(allDestinations, !is.na(state)), aes(x = state, y = n))
+g + geom_bar(stat = "identity") + coord_flip() + labs(title = "States to which All Southwest Flights Tend to Fly")
```
rudeboybert: For printing tables, consider:

* the `kable()` function from the knitr package. Converts a table to markdown table.
* [DT](https://rstudio.github.io/DT/) package

For example, try out

```r
library(DT)
library(knitr)
kable(iris)
datatable(iris)
```

HW-1.Rmd
```diff
@@ -52,54 +56,148 @@ Please indicate
...
+delay_weather <- flights %>%
+  select(date, hour, dep_delay) %>%
+  left_join(., weather_patterns, by=c("date", "hour"))
+
+delay_weather %>%
+  filter(!is.na(events)) %>% 
+  group_by(events) %>%
+  summarise(median_departure_delay = median(dep_delay)) %>%
+  arrange(desc(median_departure_delay))
+```
```
rudeboybert: While this does answer the question, it only presents first order notions of center (i.e. the median). One objective I have in this class is to get students thinking 2nd order notions of spread. i.e. the variability in the data.

Boxplots with categorical variables as the x aesthetic give a nice visualization of the entire distribution, that makes it easier to compare across levels i.e. with a straight line.

HW-1.Rmd
```diff
@@ -107,6 +205,36 @@ I want to know what proportionately regions (NE, south, west, midwest) each
...
+
+# Use those table to compute the proportion of flights per region per carrier for July.
+prop_flightsPerRegion <- flightsPerRegion_carrier %>%
+  left_join(., totalFlights_carrier, by = "carrier") %>%
+  mutate(proportionFlights = flights_perRegion/total_flights) %>% 
+  select(carrier, region, proportionFlights) %>%
+  arrange(carrier, desc(proportionFlights))
+
+as.data.frame(prop_flightsPerRegion)
 ```
```
rudeboybert: I would visualize this, or at the very least, not present the proportions with such high degree of precision. i.e. round to 3 or 4 significant digits. This is called pseudo-precision that only junks up the chart.

Take a look at this [example](https://github.com/Middlebury-Data-Science/HW-1/pull/3/files#r54886061) here using `geom_bar(stat="identity", position="fill")`

rudeboybert:
Overall looks good. One best practice is to restrict lines of code/text to be no more than 80 characters long. This harks back to the days of smaller monitors and dot matrix printers, but its still used now. You can set guide-lines in RStudio via Tools -> Global Options -> Code -> Display.

Can you let me have feedback about this feedback mechanism. So far this has made grading FAR more pleasant than downloading files from Moodle.

rkclement:
Thanks for the feedback. The tips on best practices are especially helpful, as are the tips on how to best present the analysis. I think this method for feedback is great! Keeps you from having to download files, and allows you to easily link to examples in other submissions, which is very helpful. It would be great to get all the feedback emails as a digest, rather than individually, but that's a minor quibble (at least I don't think that's possible, from poking around in my account settings...) 

Overall, works great!

PR #7, opened 2016-03-02
Delaney Moran

DelaneyMoran:


HW-1.Rmd
```diff
@@ -1,44 +1,45 @@
...
 ## Admistrative:
 
 Please indicate
 
-* Who you collaborated with:
-* Roughly how much time you spent on this HW:
-* What gave you the most trouble:
+* Who you collaborated with: Christian, Mohammed
+* Roughly how much time you spent on this HW: 7h
+* What gave you the most trouble: error messages/syntax
```
rudeboybert: yeah, there's no magic pill to make this any easier unfortunately.

HW-1.Rmd
```diff
@@ -54,6 +55,26 @@ Please indicate
...
+total_flights_day <- group_by(flights, date) %>% 
+  tally() %>% 
+  rename(total_flights = n)
+
+set1 <- full_join(total_flights_day, number_delayed, by = "date") 
+
+set1<- mutate(set1, prop = (n_delayed)/(total_flights))
+
+
+ggplot(data = set1, aes(x = date, y = prop)) + geom_bar(stat = "identity") + ylab("proportion of flights delayed more than 30min") + ggtitle("Proportion of Delayed Flights Per Day")
```
rudeboybert: For this, why `geom_bar()` and not `geom_point` or `geom_line`? While using bar's are not wrong, in a situation like this where there is a high density of values on the x axis, using bars just adds a lot of ink to the page. [Edward Tufte](http://www.edwardtufte.com/tufte/) has a rule of thumb of the information/ink ratio: maximize information while using minimal ink.

While not a hard a fast rule, I typically only use bars for displaying notions of count/frequency/density of one variable, not the relationship between two variables.

HW-1.Rmd
```diff
@@ -54,6 +55,26 @@ Please indicate
...
+  
+
+ggplot(data = set1, aes(x = date, y = prop)) + geom_smooth() + ylab("proportion of flights delayed > 30min") + ggtitle("Season Trends in Flight Delays")
```
rudeboybert: great GREAT **GREAT** idea using the smoother here. it separates out the "signal from the noise", which is incidentally the title of a [book](http://www.amazon.com/The-Signal-Noise-Predictions-Fail-but/dp/0143125087) by Nate Silver, a famous political analyst.

However, why plot it separately?  Remember, `ggplot2` works in layers that you can add things piece-by-piece.  So using your code, I suggest doing:

```r
ggplot(data = set1, aes(x = date, y = prop)) + geom_bar(stat = "identity") + 
ylab("proportion of flights delayed more than 30min") + 
ggtitle("Proportion of Delayed Flights Per Day") +
geom_smooth()
```
together.

HW-1.Rmd
```diff
@@ -67,8 +88,28 @@ they tend to have more room.  Which airlines should these people favor?
...
+
+final2 <- full_join(total_flights_carrier, set2, by = "carrier") %>% 
+  mutate(prop = num_pre1980/total_flights_carrier)
+
+
+ggplot(data = final2, aes(x = carrier, y = prop)) + geom_bar(stat = "identity")
+
+
+
+```
```
rudeboybert: perhaps you haven't had time yet, but I really have no idea what this is a plot of. Even if you don't have time to write comments (which you should be doing on the fly), there should be at least a title or some text.

Also, see a similar remark I made for Andrew at the top of this [page](https://github.com/Middlebury-Data-Science/HW-1/pull/3#discussion_r54884817)

HW-1.Rmd
```diff
@@ -82,6 +123,30 @@ they tend to have more room.  Which airlines should these people favor?
...
+
+all_flights_states <- question3 %>% 
+  group_by(state) %>%
+  filter(!is.na(state)) %>% 
+  tally() %>% 
+  rename(all_flights_by_state = n)
+
+
+ggplot(all_flights_states, aes(x = state, y = all_flights_by_state)) + geom_bar(stat = "identity") + ylab("Number of Flights")
+#add title
```
rudeboybert: I think I already shared the relevant link, but yeah, reordering the bars here is almost essential. When dealing with state data, this is known as the "ALABAMA FIRST" phenomenon.

And yes, as you say, add an informative title!

HW-1.Rmd
```diff
@@ -82,6 +123,30 @@ they tend to have more room.  Which airlines should these people favor?
...
+
+unique_flights <- question3 %>%
+  filter(!is.na(state)) %>% 
+  distinct(dest) %>% 
+  group_by(state) %>% 
+  tally() %>% 
+  rename(unique_dest = n)
+
+ggplot(unique_flights, aes(x = state, y = unique_dest)) + geom_bar(stat = "identity")
```
rudeboybert: something is not right about this plot. I don't delve deep into code at the Feedback stage, so I won't debug this for you, but you should look into this.

rudeboybert:
Overall, things are looking like they're on the right track. A couple of general comments:

* For code legibility's sake, please try to keep all lines 80 characters or less. You can have RStudio set a guide-line by going to Tools -> Global Options -> Code -> Display

DelaneyMoran:
Submission

HW-1.Rmd
```diff
@@ -53,11 +50,29 @@ Please indicate
...
+  group_by(date) %>% 
+  tally() %>% 
+  rename(n_delayed = n)
 
-```
+total_flights_day <- group_by(flights, date) %>% 
+  tally() %>% 
+  rename(total_flights = n)
+
+set1 <- full_join(total_flights_day, number_delayed, by = "date")
```
rudeboybert: You could've done this with a single data frame, by creating a vector `is_delayed` that is =1 if `delay>=30`, then `group_by(day) %>% summarise(prop = mean(is_delayed))`

HW-1.Rmd
```diff
@@ -67,11 +82,29 @@ they tend to have more room.  Which airlines should these people favor?
...
 
+plot2 <- ggplot(set2, aes(x = carrier, y = year, color = as.factor(average))) + 
+  geom_point(size = 3) + coord_flip() + xlab("Carrier") + ylab("Year") +
+  scale_fill_manual(breaks = c(0, 1), labels = c("All Planes", "Average Year")) + ggtitle("Age of Planes by Carrier")
+plot2
+
+
+```
+
+Each red dot represents a plane according to its year, and gives a sense of the range of options. The blue dots represent the average year of planes for each carrier. (The legend on the side should read: "All Planes" and "Average Year" instead of 0 and 1, but I did not have time to debug why that did not work). I also would have liked to arrange them in order by average year, and tried to fiddle around with that too, but couldn't get it.
```
rudeboybert: The best way to create a new variable with the desired legend names before you run the ggplot2. For example

```r
set2 <- set2 %>% 
  mutate(
    average_2 = ifelse(average==0, "All Planes", "Average Year"),
    average_2 = as.factor(average_2)
  )
```

HW-1.Rmd
```diff
@@ -67,11 +82,29 @@ they tend to have more room.  Which airlines should these people favor?
...
+  group_by(carrier) %>% 
+  mutate(average = 0)
 
+avg <- aggregate(year ~ carrier, set2, mean) %>% 
+  mutate(average = 1)
 
+set2 <- bind_rows(set2, avg)
 
+plot2 <- ggplot(set2, aes(x = carrier, y = year, color = as.factor(average))) + 
+  geom_point(size = 3) + coord_flip() + xlab("Carrier") + ylab("Year") +
```
rudeboybert: The problem with using a `geom_point()` here, is that a lot of points will be superimposed on top of each other. Use `geom_jitter()` so that the points get spread out a little, or better yet, use `geom_boxplot()` here, that way you don't show all the point, but summary stats of the points: 1st quartile, median, etc.

HW-1.Rmd
```diff
@@ -94,11 +165,51 @@ they tend to have more room.  Which airlines should these people favor?
...
+  geom_line(stat = "identity") +
+  geom_vline(xintercept = 32, col="red") +
+  ylab("Proportion of Flights Delayed > 30min") +
+  ggtitle("Delayed Flights by Temperature")
 
-```
 
+```
 
+The top plot indicates that Freezing Rain (and "Light Freezing Rain") are associated with the longest departure delays. The red line indicates the baseline average delay associated with "Clear" weather.It would have been interesting to plot the proportion increase in delay time by each condition, rather than just the average delay.
```
rudeboybert: I :sparkling_heart: how you established a baseline. That's awesome. What do any of these freaky weather patterns mean? Like, "I got 17/25 on a test" Without some notion of the baseline distribution, that statement is near meaningless.

The only thing is while you present notions of center, how much variability is there? like, light freezing rain has an average of about 75, but were they all 75?  were some 30 and others 100? One goal I have is to get you all to start thinking beyond just means, but also spread.

rudeboybert:
**OVERALL COMMENTS**. good job overall. I really liked the vertical and horizontal lines you drew for Q4. There is something known as the info/ink ratio.  i.e. maximize info given minimal ink. The ink needed to draw a line is minimal, but the gain in insight is great, so totally worth it.

Overall coding style is good, but when building ggplot2, please add each element on its own separate line. I find it easier to see all the components in one shot that way.

```r
# like this
ggplot() + 
   xlab("blah") +
   ylab("yo") + 
   ggtitle("hello world")

# and not this
ggplot() + xlab("blah") + ylab("yo") + ggtitle("hello world")
```

My feedback is best viewed in [Split Screen Mode](https://github.com/Middlebury-Data-Science/HW-1/pull/7/files)

PR #8, opened 2016-03-02
Enrique Dupleich

edupleich:
First submission of HW1


HW-1.Rmd
```diff
@@ -1,102 +1,153 @@
...
 
 ```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
+num_flights_per_day <- flights %>% 
+  group_by(date) %>% 
+  tally() %>% 
+  rename(num_flights = n)
 
-```
+num_delayed_flights_per_day <-  flights %>% 
+  filter(dep_delay>30 | arr_delay>30) %>%
```
rudeboybert: I would indicate in your write-up that you filter for both departure and arrival delayed by more than 30. Since this dataset focuses only on outbound flights from houston, they would assume only depature delays.

HW-1.Rmd
```diff
@@ -1,102 +1,153 @@
...
+delayed_flights <- left_join(num_flights_per_day,num_delayed_flights_per_day,by="date") %>% 
+  mutate(prop = round(num_delayed_flights/num_flights,3))
 
+ggplot(data=delayed_flights,aes(x=date,y=prop))+
+  geom_line()
+
+```
+* Which seasons did we tend to see the most and least delays of > 30 minutes.
+
+From the previous data frame we can take all delayed flights by more then 30 min and group them by season after some modifications to the original date metrics. We can observe that the number of delays does not differ considerably by season.
```
rudeboybert: Are you sure?  The data is very noisy. Think if there is a way to pick out the "signal from the noise."

HW-1.Rmd
```diff
@@ -1,102 +1,153 @@
...
+by_season <- delayed_flights
+by_season$date <- month(by_season$date)
+by_season$date[by_season$date==1|by_season$date==2|by_season$date==12] <- "Winter"
+by_season$date[by_season$date==3|by_season$date==4|by_season$date==5] <- "Spring"
+by_season$date[by_season$date==6|by_season$date==7|by_season$date==8] <- "Summer"
+by_season$date[by_season$date==9|by_season$date==10|by_season$date==11] <- "Autumn"
 
+by_season %>% rename(season=date) %>% 
+  group_by(season) %>% 
+  tally()
```
rudeboybert: I know without looking at this code that your answer is wrong. Note: 91+92+92+90 = 365.

You've fallen into a very serious trap here that people who are more mathematically/computationally oriented fall into when working with "real data" in a "real context". They think too much about "how to get a result" and not enough on "what the result means". Even a 5 second sanity check of these results would show the error. This sounds like really harsh medicine, but I myself have gotten burned in the past like this, and it cost me. 

Getting an answer is the easy part, it's interpreting it and adding value as an analyst that's the hard part.

HW-1.Rmd
```diff
@@ -1,102 +1,153 @@
...
 
 ```{r, echo=FALSE, fig.width=12, fig.height=6}
-
+pref_old_planes <- inner_join(planes,flights,by="plane") %>% 
+  filter(!is.na(year)) %>% 
+  select(plane,year,carrier) %>% 
+  group_by(carrier) %>% 
+  summarise(year_mean=round(mean(year),0),num_distinct_planes_by_carrier=n_distinct(plane)) %>% 
+  arrange(year_mean)
+pref_old_planes
```
rudeboybert: It's really good that you present the average age of the planes AND the number of planes. b/c yes, who cares if the average year is 1982 if a carrier only has 2 planes.

However, take AA for example. The avg year manufactured is 1987. Were all its planes made in 1987? Or did you have some in 1957 and some in 2016. As a traveller wanting to pick a carrier, maybe this isn't enough information.

HW-1.Rmd
```diff
@@ -1,102 +1,153 @@
...
 
+I filtered all the unique flights that had identifiers ending with "SW" and then performed a n_distinct summarization
 ```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
-
+flights_with_airports <- left_join(flights,airports, by=c("dest"="iata"))
+listed_SW_states_tendency <- flights_with_airports %>% 
+  filter(substr(plane,5,6)=="SW") %>% 
+  group_by(state) %>% 
+  summarise(n_distinct(plane))
+listed_SW_states_tendency
```
rudeboybert: You only list the first 10 states here. Also, why does this table have 29 rows, but the next only have 20 rows?  Shouldn't the listed states be the same?

HW-1.Rmd
```diff
@@ -1,102 +1,153 @@
...
-
-
+all_SW_states_tendency <-  flights_with_airports%>% 
+  filter(carrier=="WN") %>% 
+  select(carrier,dest,state) %>% 
+  group_by(state) %>% 
+  tally() %>% 
+  arrange(desc(n))
+all_SW_states_tendency
+```
```
rudeboybert: Here you sort by number, which is good. why not a visualization? people absorb information much better visually.

HW-1.Rmd
```diff
@@ -108,5 +159,19 @@ carrier flies to from Houston in the month of July.  Consider the `month()`
...
+num_per_carrier_region <- left_join(flights_with_airports,states,by="state") %>% 
+  group_by(carrier,region) %>% 
+  tally() %>% 
+  rename(num_carrier_region=n) %>% 
+  arrange(carrier)
+
+region_carrier_prop_july <- left_join(num_per_carrier_region, num_per_carrier, by="carrier") %>% 
+  mutate(prop = round(num_carrier_region/num_flights,3))
+region_carrier_prop_july
 ```
```
rudeboybert: Again, this table is of almost no use to someone who needs to take this information and use it. You

* only print the first 10 rows
* even if you did print all the rows, you'd have 37x5 values to consider at once, which really stretches the cognitive bandwidth of people. again, this is where visualizations are key.

HW-1.Rmd
```diff
@@ -1,102 +1,153 @@
...
+all_SW_states_tendency
+```
 
 ## Question 4:
 
-What weather patterns are associated with the biggest departure delays?
+* What weather patterns are associated with the biggest departure delays?
 
-```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
+For this question I removed the columns precip, gust_speed and events because they hosted mostly NA values. Then I filtered the flights that were delayed by more than 30 min (the median of all delayed flights was 10 minutes). Finally I grouped the average of the conditions by conditions reported.
```
rudeboybert: You present averages, but again, for example for `Freezing Rain`, the mean `dep_delay` is 219.2.  Was it 219.2 everytime?

rudeboybert:
Hi Enrique, I'm glad you submitted a feedback request, b/c if this was your submission, I would give you a B-, but only if I were feeling generous.  

The fact that you presented a table in Q1 that essentially only counted the # of days in each season indicates to me that you put almost no thought into interpreting the results.

This is a class on Data Science, not on coding/R. Data science isn't just about the "data" i.e. the coding/statistics/mathematics, but also about the "science" i.e. interpretation/analysis/presentation of conclusions. If you presented this analysis to someone who worked in the airline industry, I'm all but certain they would throw it out. There is very little in this report can use.

Do not take this as an insult on your character (b/c based on our limited interactions you strike me as a very intelligent deep thinker) but rather as a wakeup call. While being a deep thinker is a necessary condition for being a good data scientist, it is not a sufficient condition.

As I said earlier, when I started out my career as a data scientist/analyst, I fell into similar traps: focusing too much on the "data" (how do I get compute result) and not enough on the "science" part (what does the result mean). Hopefully I can help you avoid doing the same if you do take a data science path with your career.

HW-1.Rmd
```diff
@@ -1,105 +1,207 @@
...
+* Which seasons did we tend to see the most and least delays of > 30 minutes.
 
 
+```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
+by_season <- delayed_flights
+by_season$month <- month(by_season$date)
+by_season$season[by_season$month==1|by_season$month==2|by_season$month==12] <- "Winter"
+by_season$season[by_season$month==3|by_season$month==4|by_season$month==5] <- "Spring"
+by_season$season[by_season$month==6|by_season$month==7|by_season$month==8] <- "Summer"
+by_season$season[by_season$month==9|by_season$month==10|by_season$month==11] <- "Autumn"
```
rudeboybert: Another way to do this that would be cleaner is using the `ifelse()` command. ie `ifelse(month %in% c(1,2,3), "Spring", season)`

HW-1.Rmd
```diff
@@ -1,105 +1,207 @@
...
+
+by_season  %>% 
+  group_by(season) %>% 
+  summarize_each(funs(mean(.),sd(.)),num_delayed_flights,prop)
```
rudeboybert: Reading raw R table outputs are very jarring. Consider the
* the `kable()` command in the `knitr` package
* the DT package: https://rstudio.github.io/DT/

Also, it would be best to rename the variables at the final step to have column names in english, rather than computer code.

HW-1.Rmd
```diff
@@ -1,105 +1,207 @@
...
+  group_by(season) %>% 
+  summarize_each(funs(mean(.),sd(.)),num_delayed_flights,prop) 
+
+ggplot(data=by_season,aes(x=date,y=prop))+
+  geom_line()+
+  geom_smooth()+
+  facet_wrap(~season, scales="free")
 
+```
+From the previous data frame we can take all delayed flights by more then 30 min and group them by season after some modifications to the original date values. By taking the average proportion of delayed flights for each season we can observe that the season with the net most delays is Spring; proportionately it is Summer while the season with the least delays over 30min is Autumn. Below is the same plot as above but divided by sections. I couldn't figure out how to set the x-scale individually, therefore the plot for Winter is rather useless.
```
rudeboybert: Yep. I didn't know about `scales="free"`. Also, for the winter plot, yeah, these kinds of things are annoying: the calendar year does not line up with the season schedule.

HW-1.Rmd
```diff
@@ -1,105 +1,207 @@
...
+  select(plane,year,carrier)
+
+pref_old_planes %>% 
+  group_by(carrier) %>% 
+  summarise(year_mean=round(mean(year),0),
+            num_diff_planes_carrier=n_distinct(plane),
+            max(year),
+            min(year),
+            sd=round(var(year),1)) %>% 
+  arrange(year_mean,sd)
```
rudeboybert: Again, see earlier comment about raw table outputs being jarring.

HW-1.Rmd
```diff
@@ -1,105 +1,207 @@
...
+  group_by(date) %>% 
+  tally() %>% 
+  rename(num_delayed_flights = n)
+
+delayed_flights <- left_join(num_flights_per_day,num_delayed_flights_per_day,by="date") %>% 
+  mutate(prop = round(num_delayed_flights/num_flights,3))
 
+ggplot(data=delayed_flights,aes(x=date,y=prop))+
+  geom_line() +
+  geom_smooth()
```
rudeboybert: Please put thought into the labels and titles. Often times what gets passed around in work settings is just the graphic, so the plots need to be as self-explanatory as possible.

HW-1.Rmd
```diff
@@ -1,105 +1,207 @@
...
+pref_old_planes %>% 
+  group_by(carrier) %>% 
+  summarise(year_mean=round(mean(year),0),
+            num_diff_planes_carrier=n_distinct(plane),
+            max(year),
+            min(year),
+            sd=round(var(year),1)) %>% 
+  arrange(year_mean,sd)
+```
+MQ's planes are on average the oldest, but they only have 2 planes. AA and DL have many more and they average planes built in 1987 and 1990 respectively. However, there is a lot of variance in the data. With the aid of a box plot we can confirm that AA has the greatest number of old planes across their fleet, despite the high SD of 37.8
```
rudeboybert: Good job pointing out the sample size. It does matter in the interpretation!

HW-1.Rmd
```diff
@@ -1,105 +1,207 @@
...
-
 
+For this part of the question I filtered flights by carrier == "WN" which Google claims is the  Southwest Airlines AVIACSA 2 letter code and then performed a n_distinct summarization
 ```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
+flights_with_airports <- left_join(flights,airports, by=c("dest"="iata"))
+listed_SW_states_tendency <- flights_with_airports %>% 
+  filter(carrier=="WN") %>% 
+  na.omit() %>% 
+  group_by(state) %>% 
+  summarise(uniq=n_distinct(plane)) %>%
```
rudeboybert: had to `n_distinct(flights)` here. we are looking for unique flight patterns, not unique airplanes used.

Also, if dropping missing values, one should mention this in the analysis. If the data are missing for a *systematic reason*, this will affect your conclusions.

HW-1.Rmd
```diff
@@ -1,105 +1,207 @@
...
+  na.omit() %>% 
+  select(carrier,dest,state) %>% 
+  group_by(state) %>% 
+  tally() %>% 
+  arrange(desc(n))
+all_SW_states_tendency
+```
+Below are bar plots for unique WN flights and all WN flights
+```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
+ggplot(listed_SW_states_tendency,aes(reorder(state,uniq),uniq))+geom_bar(stat="identity")
```
rudeboybert: Wow! I didn't know about the `reorder` command! I'm still learning new R commands today! Good job not falling into the ALABAMA FIRST problem. Also, see note about axes labels and titles.

HW-1.Rmd
```diff
@@ -1,105 +1,207 @@
...
+dep_delay_flights <- inner_join(x,y,by=c("date","hour")) %>% 
+  na.omit() %>% 
+  arrange(desc(dep_delay))
+
+most_delaying_conditions <- dep_delay_flights %>% 
+  na.omit() %>%   
+  group_by(conditions) %>% 
+  summarise(avg_delay=mean(dep_delay),sd_delay=sd(dep_delay),avg_temp=mean(temp),n()) %>% 
+  arrange(desc(avg_delay))
+most_delaying_conditions
```
rudeboybert: Again, good job showing the sample size using the `n()` function. Any explanations as to why Thunderstorms have lower mean delays than Clear?

HW-1.Rmd
```diff
@@ -1,105 +1,207 @@
...
 
+There seems to be an increase in delay times when pressure is around 30
+```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
 
+ggplot(dep_delay_flights,aes(pressure,dep_delay))+geom_jitter()
+```
+There seems to be an increase in the delay times when the wind comes from the south and when the other wind direction metric is between 100 and 200 and beyond 300.
+```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
 
+ggplot(dep_delay_flights,aes(wind_dir,dep_delay))+geom_jitter()
```
rudeboybert: I think boxplots would've been better here.

HW-1.Rmd
```diff
@@ -1,105 +1,207 @@
...
+Temperature
+```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
+ggplot(dep_delay_flights,aes(temp,dep_delay))+geom_jitter()
+```
+Pressure
 
+There seems to be an increase in delay times when pressure is around 30
+```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
 
+ggplot(dep_delay_flights,aes(pressure,dep_delay))+geom_jitter()
```
rudeboybert: Adding a regression line here would really help pick out the signal from the noise.

HW-1.Rmd
```diff
@@ -1,105 +1,207 @@
...
+There seems to be an increase in the delay times when the wind comes from the south and when the other wind direction metric is between 100 and 200 and beyond 300.
+```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
 
+ggplot(dep_delay_flights,aes(wind_dir,dep_delay))+geom_jitter()
+ggplot(dep_delay_flights,aes(wind_dir2,dep_delay))+geom_jitter()
+```
+Low wind speeds also increase delay times.
+```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
+ggplot(dep_delay_flights,aes(wind_speed,dep_delay))+geom_jitter()
+```
```
rudeboybert: Overall, you have to watch your language in your conclusions. For example "Low wind speeds also increase delay times." are we sure about this? did we run an experiment?  All we can say is "Low wind speeds are associated with increased delay times. "

HW-1.Rmd
```diff
@@ -107,6 +209,29 @@ I want to know what proportionately regions (NE, south, west, midwest) each
...
 
-```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
+Below is a table with the region-carrier proportions and a bar plot for each carrier:
 
+```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
+flights_with_airports <- flights_with_airports %>% filter(month(date,label =TRUE)=="Jul")
+num_per_carrier <-  flights_with_airports%>% 
+  group_by(carrier) %>% 
+  tally() %>% 
+  rename(num_flights=n)
```
rudeboybert: At a certain point, tables that are too big are not worth showing. While one could argue this one isn't TOOO big, its getting there.

HW-1.Rmd
```diff
@@ -107,6 +209,29 @@ I want to know what proportionately regions (NE, south, west, midwest) each
...
+region_carrier_prop_july <- left_join(num_per_carrier_region, num_per_carrier, by="carrier") %>% 
+  na.omit() %>% 
+  mutate(prop = round(num_carrier_region/num_flights,3))
+as.data.frame(region_carrier_prop_july)
+
+ggplot(region_carrier_prop_july,aes(region,prop))+
+  geom_bar(position="dodge",
+           stat="identity",
+           aes(fill=region))+
+  facet_wrap(~carrier)
```
rudeboybert: Good job showing proportions instead of raw counts. But why not use `position="stack"`. i.e.

```r
ggplot(region_carrier_prop_july,aes(carrier,prop))+
  geom_bar(position="stack",
           stat="identity",
           aes(fill=region))
```

rudeboybert:
**OVERALL COMMENTS**: As Donald Trump and Bernie Sanders would say, this was a YOOOOOOOOGE improvement.

My feedback is best view in [split screen mode](https://github.com/Middlebury-Data-Science/HW-1/pull/8/files)

PR #12, opened 2016-03-09
Carter Merenstein

cmerenstein:
It requires ggmap and TD, but will install both if you don't have them.

### DIFF
--- a/HW-1.Rmd
+++ b/HW-1.Rmd
@@ -1,16 +1,20 @@
 ---
 title: "MATH 216 Homework 1"
-author: "WRITE YOUR NAME HERE"
+author: "Carter Merenstein"
 output: html_document
 ---
 
 ```{r, echo=FALSE, message=FALSE}
 # DO NOT EDIT THIS SECTION
 # Load packages
+
+if("ggmap" %in% rownames(installed.packages()) == FALSE) {install.packages("ggmap")}
+
 suppressPackageStartupMessages(library(ggplot2))
 suppressPackageStartupMessages(library(dplyr))
 suppressPackageStartupMessages(library(tidyr))
 suppressPackageStartupMessages(library(lubridate))
+suppressPackageStartupMessages(library(ggmap))
 ```
 ```{r, echo=FALSE, message=FALSE, cache=TRUE}
 # Load data sets. Note this assumes this .Rmd files is in the same directory as
@@ -39,10 +43,10 @@ states <- read.csv("data/states.csv", stringsAsFactors = FALSE) %>%
 
 Please indicate
 
-* Who you collaborated with:
-* Roughly how much time you spent on this HW:
-* What gave you the most trouble:
-* Any comments you have:
+* Who you collaborated with: no one
+* Roughly how much time you spent on this HW: 6 hr
+* What gave you the most trouble: r syntax
+* Any comments you have: Learning R is a lot more frustrating than learning other programming languages. I feel like this assignment would have taken 2 hours in any other language but I lost a lot of time figureing out the differences between things like sort(), order(), and arrange(), or n(), count(), tally() and sum(). There also don't seem to be really good documentations for a lot of things. The help() function is not nearly complete as something like man() or the java API. For example, I want a list of all possible options for the stat argument in ggplot, but nothing seeems to just list them.
 
 
 
@@ -60,7 +64,7 @@ q1
 
 q1_delay = q1 %>% mutate(arr_delay = ifelse(is.na(arr_delay),31, arr_delay))
 #cancled flights are really delayed for more than 30 minues
-q1_delay
+#q1_delay
 q1_delay = na.omit(q1_delay) #now only removes values that are na for both
 #View(q1_delay)
 
@@ -68,12 +72,12 @@ q1_delay = na.omit(q1_delay) #now only removes values that are na for both
 delays = group_by(q1_delay, date) %>% summarize(del = sum(arr_delay > 30))
 total = group_by(q1_delay, date) %>% summarize(del = n())
 
-delays
-total
+#delays
+#total
 
 proportion = full_join(total, delays, by="date")
 proportion = mutate(proportion, prop = del.y / del.x)
-proportion
+#proportion
 
 p1 = ggplot(proportion, aes(x = date, y = prop))
 p1 = p1 + geom_point()
@@ -118,11 +122,14 @@ SW = filter(flights, carrier == "WN")
 SW = inner_join(airports, SW, by = c("iata"="dest"))
 SW = inner_join(states, SW, by = "state")
 
-px = ggplot(SW, aes(x = long, y = lat, color = state)) + geom_point()
-px ## this is just for curiosity's sake, but it looks cool! If only I could figure out how to put a real map under it
-
-q3 = select(SW, `state`, `date`, `iata`)
+q3 = select(SW, `state`, `date`, `iata`, `lat`, `long`)
 q3a  = group_by(q3, state) %>% summarize(Flights = n()) ## all flights
+
+q3_map = inner_join(q3a, q3, by = "state")
+
+p_map = qmap("USA", zoom = 4, color = "bw")
+p_map = p_map + geom_point(data = q3_map, aes(x = long, y = lat, fill = state, color = "black", size = Flights), pch=21) + scale_shape_identity()
+p_map
 
 q3a = arrange(q3a, Flights)
 q3a
@@ -188,8 +195,8 @@ q5b = q5 %>% group_by(carrier, region) %>% summarize(region_total = n())
 q5 = inner_join(q5a, q5b, by = "carrier") %>% mutate(percent_in_region = (region_total/carrier_total)*100)
 q5
 
-p5 = ggplot(q5, aes(x = carrier, y = percent_in_region, fill = region)) + geom_bar(stat = "identity")
-p5 = p5 + scale_fill_brewer(palette = "Spectral")
+p5 = ggplot(q5, aes(x = carrier, y = percent_in_region, fill = factor(region))) + geom_bar(stat = "identity", color = "black")
+p5 = p5 + scale_fill_brewer(palette = "Pastel2")
 p5
 
 
--- a/HW-1.Rmd
+++ b/HW-1.Rmd
@@ -4,29 +4,32 @@ author: "WRITE YOUR NAME HERE"
 output: html_document
 ---
 
-```{r, echo=FALSE, message=FALSE, cache=TRUE}
+```{r, echo=FALSE, message=FALSE}
 # DO NOT EDIT THIS SECTION
-
 # Load packages
 suppressPackageStartupMessages(library(ggplot2))
 suppressPackageStartupMessages(library(dplyr))
+suppressPackageStartupMessages(library(tidyr))
 suppressPackageStartupMessages(library(lubridate))
-
+```
+```{r, echo=FALSE, message=FALSE, cache=TRUE}
 # Load data sets. Note this assumes this .Rmd files is in the same directory as
 # the .csv files.
 flights <- read.csv("data/flights.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df() %>%
-  mutate(date=as.Date(date))
+ tbl_df() %>%
+ mutate(date=as.Date(date))
 weather <- read.csv("data/weather.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df() %>%
-  mutate(date=as.Date(date))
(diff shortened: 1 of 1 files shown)


rudeboybert:
Got it. However, the auto-installation of the packages didn't work on my end. So I commented them out, installed them manually, and everything Knited. IMO its not a good idea for force installation of packages onto people; rather make it clear that they need to install whatever packages, and let the user decide if they want to.

HW-1.Rmd
```diff
@@ -1,45 +1,58 @@
...
 Please indicate
 
-* Who you collaborated with:
-* Roughly how much time you spent on this HW:
-* What gave you the most trouble:
-* Any comments you have:
+* Who you collaborated with: no one
+* Roughly how much time you spent on this HW: 6-8 hr
+* What gave you the most trouble: r syntax
+* Any comments you have: Learning R is a lot more frustrating than learning other programming languages. I feel like this assignment would have taken 2 hours in any other language but I lost a lot of time figureing out the differences between things like sort(), order(), and arrange(), or n(), count(), tally() and sum(). There also don't seem to be really good documentations for a lot of things. The help() function is not nearly complete as something like man() or the java API. For example, I want a list of all possible options for the stat argument in ggplot, but nothing seeems to just list them.
```
rudeboybert: * when I read help files, I typically look at the examples at the bottom. I agree, just looking at the argument list isn't always the most helpful
* ggplot is a tricky package to learn, agreed. I don't use the help files as much as I use examples from the ggplot 2 page: http://docs.ggplot2.org/current/

HW-1.Rmd
```diff
@@ -51,10 +64,37 @@ Please indicate
     + the x-axis should be some notion of time
     + the y-axis should be the proportion.
 * Which seasons did we tend to see the most and least delays of > 30 minutes.
-
 ```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
 
+q1 = select(flights, `date`, `arr_delay`, `cancelled`)
```
rudeboybert: Nitpick: the R community generally uses `<-` for assignment rather than `=`. While not a big deal, its better to stick with conventions used by most users.

HW-1.Rmd
```diff
@@ -51,10 +64,37 @@ Please indicate
...
+
+delays = group_by(q1_delay, date) %>% summarize(del = sum(arr_delay > 30))
+total = group_by(q1_delay, date) %>% summarize(del = n())
+
+#delays
+#total
+
+proportion = full_join(total, delays, by="date")
+proportion = mutate(proportion, `Percent Delayed` = round((del.y / del.x)*100, digits = 2))
+proportion = rename(proportion, Delayed = del.y, Total = del.x)
```
rudeboybert: if using dplyr commands on the same object, using the pipe operator necessitates less repetition of typing and emphasizes the step-by-step manipulations on the object.

HW-1.Rmd
```diff
@@ -51,10 +64,37 @@ Please indicate
...
+
+#proportion
+
+datatable(proportion, options = list(searching = FALSE, pageLength = 20, lengthMenu = c(5, 10, 20, 50)))
```
rudeboybert: Nice job digging up this function, but I'm not sure what value showing all 365 values adds on top of what's already presented in the graph below.

HW-1.Rmd
```diff
@@ -51,10 +64,37 @@ Please indicate
...
+
+p1 = ggplot(proportion, aes(x = date, y = `Percent Delayed`))
+p1 = p1 + geom_point()
```
rudeboybert: adding a `geom_smooth()` here would really allow you to pick out the signal out of the noise.

HW-1.Rmd
```diff
@@ -67,12 +107,18 @@ they tend to have more room.  Which airlines should these people favor?
...
 
-```
-
-
 
+q2 = inner_join(planes, flights, by = "plane")
+q2 = select(q2, `plane`, `year`, `carrier`)
+#q2
 
+p2 = ggplot(q2, aes(x = year))
```
rudeboybert: unless you are going to recycle the base layer of a ggplot for other plots, its better practice to list out all the elements at once. i.e.
```r
p2 = ggplot(q2, aes(x = year)) + 
    geom_histogram() + 
    facet_wrap(~carrier)
```

HW-1.Rmd
```diff
@@ -67,12 +107,18 @@ they tend to have more room.  Which airlines should these people favor?
...
+p2 = p2 + geom_histogram()
```
rudeboybert: This histogram is of limited use outside of Continental, XE and Southwest: I can't see whats going on with DL, and many of the other airlines b/c the counts of the biggest 3 are blowing up the y scale. Also with histograms, you have to make comparisons between airlines along two dimensions (x and y axes).

A `geom_boxplot` alleviates this problem b/c you can make comparisons using a single line, b/c all the action in a boxplot is along one dimension. We'll go over some examples on Wednesday.

HW-1.Rmd
```diff
@@ -80,9 +126,46 @@ they tend to have more room.  Which airlines should these people favor?
 * What states did listed Southwest Airlines flights tend to fly to?
 * What states did all Southwest Airlines flights tend to fly to?
 
+```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
+
+
+SW = filter(flights, carrier == "WN")
+SW = inner_join(airports, SW, by = c("iata"="dest"))
+SW = inner_join(states, SW, by = "state")
```
rudeboybert: Again, using piping here is a little less jarring. You can target a pipe to a specific argument by using a period:

```r
SW <- filter(flights, carrier == "WN") %>%
    inner_join(airports, ., by = c("iata"="dest")) %>%
    inner_join(states, ., by = "state")
```

HW-1.Rmd
```diff
@@ -80,9 +126,46 @@ they tend to have more room.  Which airlines should these people favor?
...
+p3a = ggplot(q3a, aes(x = state, y = Flights)) + geom_bar(stat = "identity")
+p3a
+```
+> This shows all flights that Southwest makes, meaning it's skewed to routes that are repeated daily or more. Still, this demonstrates that most of Southwest's buisness out of Texas is regional, staying in Texas or the rest of the south.    
+> The graph below shows just unique flights, meaning there are some states that only get a single flight. These bar graphs show that Southwest is making most of it's total flights in Texas, which is partially a result of flying to 7 different locations, but also a factor of high frequency regional flights. One thing that also stands out is that the third most flown-to state, LA, is really just one very frequent flight.
 
 ```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
 
+q3b = select(q3, `state`, `iata`)
+q3b = distinct(q3b) %>% group_by(state) %>% summarize(`Distinct flights` = n())
```
rudeboybert: You are `distinct()` ing using the wrong variable. For example, do a `View()` of SW flights. You'll see there are multiple flight #'s that fly to ABQ and DAL (I think there were 8).  So what you needed to do here was `distinct(flight)`

HW-1.Rmd
```diff
@@ -80,9 +126,46 @@ they tend to have more room.  Which airlines should these people favor?
...
+SW = inner_join(airports, SW, by = c("iata"="dest"))
+SW = inner_join(states, SW, by = "state")
+
+q3 = select(SW, `state`, `date`, `iata`, `lat`, `long`)
+q3a  = group_by(q3, state) %>% summarize(Flights = n()) ## all flights
+
+q3a = arrange(q3a, Flights)
+#q3a
+
+p3a = ggplot(q3a, aes(x = state, y = Flights)) + geom_bar(stat = "identity")
```
rudeboybert: Your bar plots suffer from the ALABAMA FIRST phenomenon. Sorting the bars by frequency, instead of alphabetically, makes it easier to get a sense of rank. For example, looking at your plots, quickly tell me which states have the 4th and 5th most flights?

HW-1.Rmd
```diff
@@ -80,9 +126,46 @@ they tend to have more room.  Which airlines should these people favor?
...
+```
+> As clear as the bar graphs are, I think this mapping demonstrates both parts of the question more clearly. The points on the map clearly show where Southwest is flying in terms of listed fligths, and the size of the dot indicates the total number of flights to that location.
+
+
+```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=10}
+
+q3_map = inner_join(q3a, q3, by = "state")
+
+USA = qmap("USA", zoom = 4, color = "bw")
+p_map_all = USA + geom_point(data = q3_map, aes(x = long, y = lat, fill = state, size = Flights), pch=21, colour = "black") + scale_shape_identity()
```
rudeboybert: Great job going above and beyond what was required. Maps are definitely more appropriate for data like this than barplots!

HW-1.Rmd
```diff
@@ -95,6 +178,24 @@ What weather patterns are associated with the biggest departure delays?
...
+> Looking more closely at the data though, this time in a jitter plot, it becomes obvious that the drizzel effect is really just a small sample size. It also makes it clear that freezing rain isn't necessarily so bad, it just hasn't happened that often. We really don't know accuratly how freezing rain effects delays, only that the 5 times we saw it planes didn't get out on time.   
+>The jitter plot also shows some nuanced differences between otherwise similar conditions. For exmple, Rain and Scattered Clouds have similar average delays, but many more Scattered Clouds days had long delays 30 minutes or more. Looking at these data, I'd rather fly on a rainy day than a cloudy one; the average delay will be similar but the worse case scenario seems worse with scattered clouds.
+
+
+```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
+p4_conditions = ggplot(q4, aes(x = conditions, y = `dep_delay`)) + geom_point(position = position_jitter(width = 1), size = 0.25) + theme(axis.text.x = element_text(angle = 45, hjust = 1))
```
rudeboybert: This is an excellent visualization that address both the issue of
* what's the variability within each condition?
* what are the sample sizes?

HW-1.Rmd
```diff
@@ -109,4 +210,21 @@ function from the `lubridate` package.
...
+q5 = inner_join(q5a, q5b, by = "carrier") %>% mutate(percent_in_region = round((region_total/carrier_total)*100, digits = 2))
+
+datatable(q5, options = list(searching = FALSE, pageLength = 20, lengthMenu = c(5, 10, 20)))
+
+p5 = ggplot(q5, aes(x = carrier, y = percent_in_region, fill = factor(region))) + geom_bar(stat = "identity", color = "black")
+p5 = p5 + scale_fill_brewer(palette = "Pastel2")
+p5
+
 ```
+> Most airlines only fly to a small number of destinamtions, sometimes only 1 city, making their regional breakdown somewhat uninteresting. Amoung larger carriers, the overall picture is that planes from Houston mostly are headed to the south. The only major airline to buck this trend is United, which flies heavily to the West.
```
rudeboybert: I find it surprising that so few flights go to the NE. meaning if I were from houston, I would pretty much have to change flights somewhere, even to go to major cities like NY or Boston

HW-1.Rmd
```diff
@@ -109,4 +210,21 @@ function from the `lubridate` package.
...
+q5 = mutate(flights, month = month(date, label=TRUE))
+q5 = filter(q5, month=="Jul")
+q5 = inner_join(q5, airports, by=c("dest" = "iata"))
+q5 = inner_join(q5, states, by = "state")
+q5 = select(q5, `region`, `carrier`)
+q5a = q5 %>% group_by(carrier) %>% summarize(carrier_total = n()) 
+q5b = q5 %>% group_by(carrier, region) %>% summarize(region_total = n())
+q5 = inner_join(q5a, q5b, by = "carrier") %>% mutate(percent_in_region = round((region_total/carrier_total)*100, digits = 2))
+
+datatable(q5, options = list(searching = FALSE, pageLength = 20, lengthMenu = c(5, 10, 20)))
```
rudeboybert: Again, while the table is a cool feature, I'd argue its a bit redundant in light of the plot.

If however, you still want to get exact values, I suggest an interactive visualization, something along the lines of: https://plot.ly/ggplot2/geom_bar/

rudeboybert:
**OVERALL COMMENTS**: Some short comings in some of the visuals, but a solid job in others: like the map and especially the conditions jitter plot.

I can tell that you are a pretty experienced coder. However, I did find your code a bit hard to read. If you are planning on working in a collaborative environment, having code that is easy for others to understand will greatly improve the marketability of your work. Please put some thought into using piping, keeping code lines under 80 characters, and such.

My feedback is best viewed in [Split screen mode](https://github.com/Middlebury-Data-Science/HW-1/pull/12/files)

PR #14, opened 2016-03-09
Andrew Holtz

amholtz:
Homework 01 Final Submission-

Note: I some how messed up the push/ pull and commit on my original project. I ended up making a new project and uploading from there. It seemed to work- I hope this is okay.

Thanks, 
Andrew


rudeboybert:
Yeah, when you're having problems with git, that's what you do: hit the reset button. Your HTML file is knitting. I'll be going over these tomorrow.

HW-1.Rmd
```diff
@@ -52,11 +55,88 @@ Please indicate
...
+  filter(dep_delay > 30) %>% 
+  group_by(date) %>% 
+  tally() %>% 
+  rename(num_delayed_flights = n) %>% 
+  arrange(date)
+
+prop_delayed <- left_join(total_flights,delay_flights, by = "date") %>% 
+  
+  mutate(prop = round(num_delayed_flights / num_flights , 3))
+
```
rudeboybert: You could've done lines 83 thru 100 by creating a variable `is.delayed` which equals 1 if delayed >= 30 minutes, 0 otherwise, and then `group_by()` date, and then `summarise` to get the proportion.

HW-1.Rmd
```diff
@@ -52,11 +55,88 @@ Please indicate
...
+should not effect the overall conclusion. The outlier represents one particular 
+day that had many delays. This could be due to an element independent from the 
+season altogether such as logistics and holiday weekends, and therefore, should 
+not affect the conclusion. It is important to understand the limitation of 
+this figure. Fall could have many flights delayed by 29 minutes. Therefore, we 
+cannot make the conclusion that Spring and Summer have MORE delayed flights than
+Fall, but only that Spring and Summer have more flights delayed over 30 minutes 
+than Fall 2011. In conclusion, if one wished to avoid delays of 30 minutes or 
+more in 2011, one should have flown in the fall or the winter rather than the 
+spring or summer.
```
rudeboybert: This is a bit verbose, you spend a lot of time describing

* what a boxplot is (which might be appropriate given your audience)
* features of the boxplot one can already plainly see

I think the last 4-5 sentences would've sufficed.


Also your title is misleading. The units of analysis are days (the day frame has 365 rows). "Proportion of flights delayed per Day" would've been clearer.

HW-1.Rmd
```diff
@@ -52,11 +55,88 @@ Please indicate
...
+  SS <- as.Date("2011-06-21",  format = "%Y-%m-%d") # Summer Solstice
+  FE <- as.Date("2011-09-21",  format = "%Y-%m-%d") # Fall Equinox
+  
+  ifelse (DATES >= WS | DATES < SE, "Winter",
+          ifelse (DATES >= SE & DATES < SS, "Spring",
+                  ifelse (DATES >= SS & DATES < FE, "Summer", "Fall")))
+}
+  
+#This creates a table with date as one variable and the season association as the other
+date <- as.Date("2011-01-01", format = "%Y-%m-%d") + 0:364
```
rudeboybert: Unfortunately, working with dates is one of the worse part of data analyses, IMO. We'll study the `lubridate` package later, which doesn't make the task pleasant, but rather makes it less unpleasant.

HW-1.Rmd
```diff
@@ -65,48 +145,331 @@ Please indicate
...
+CO's planes is in the 2000s. One cannot chose which plane they fly on when 
+booked a flight. It is mostly chance, which plane they are seating on. They can 
+only rely on probability. Therefore, those individuals have the greatest 
+probability of securing a seat on an old plane if they fly with those airlines 
+with the lowest medians (lowest meaning oldest). Lastly an important note, to 
+make this plot, the flights that had 'na' for year of manufacture were removed.
+It is possible that these planes were too old to have their year of manufacture
+noted. It is not possible to determine the true age of these planes, so they 
+were, therefore, removed from our plot. They are, nevertheless, still important
+to note.
```
rudeboybert: Good job referencing the 1980 deregulation. This is the kind of value an analyst should add.

In your write up however, referencing more obscure airlines by their codes isn't that helpful. What are EV, F9, and YV? While in the plots due to space restrictions it makes sense to refer by code, in your write up you should write out the full airline name.

Also, don't shy from using line breaks. Large paragraphs are hard to read.

HW-1.Rmd
```diff
@@ -65,48 +145,331 @@ Please indicate
...
+  factor(repeated_flight_num$number_of_destinations, levels = repeated_flight_num$number_of_destinations[order(-repeated_flight_num$numflights_with_frequency)])
+
+#Plot 04
+z <- ggplot(data = repeated_flight_num, aes(x= number_of_destinations, y = numflights_with_frequency))
+
+z + 
+  geom_bar(stat="identity") +
+  xlab("Number of Destination Airports") +
+  ylab("Count of Flight numbers") +
+  ggtitle("Plot #04 \n The Count of Individual Flight Numbers to Multiple Destinations")
```
rudeboybert: Wow, up to 10 destinations for the same flight #? shocking indeed! good catch!

HW-1.Rmd
```diff
@@ -65,48 +145,331 @@ Please indicate
...
+  tally()
+
+state_freq$state <- 
+  factor(state_freq$state, levels = state_freq$state[order(-state_freq$n)])
+
+#Plot 03
+r <- ggplot(data = state_freq, aes(x= state, y = n)) 
+r + 
+  xlab("State") + ylab("Frequency") + 
+  ggtitle("Plot #03 \n Southwest Airline's most flied to States by flight number") +
```
rudeboybert: Another way to call this would be "flight path"

HW-1.Rmd
```diff
@@ -65,48 +145,331 @@ Please indicate
...
+100% if the air was cooled. This is way to measure moisture in the air. The 
+higher the dew point temperature the more water vapor is in the air. Therefore, 
+dew point is similar to humidity, and it relates to precipitation. Days with high
+dew point temperature could be relatively stormy or raining days. It is
+important not to forget the magnitude of each point on the plot. The darker 
+the circle, the more common that point is. The most common points are those 
+between 60ºF and 80ºF, which also have the highest delays. Therefore, it is 
+difficult to determine weather the flight delays are due to the dew point temp 
+or if the delays are due to logistical or other weather problems that just 
+happen to occur while dew point is relatively normal between 60 and 80ºF.
```
rudeboybert: In a future analyses, one might quote the confidence interval on the slope of the regression line. i.e. there is a positive relationship, but how certain are we of it?

Also your statement of "the higher the dew point temperature became, the longer the delay became" sounds a little bit like you're asserting causation. We didn't run an experiment, so we can't say this. I would word it a little more conservatively and state "increases in dew point temperature were associated with longer delays".

And nicely worded last sentence. One DEFINITELY needs to account for more potential confounders than just dew point when trying to predict delays. Again, that's why we're only making statements of association between dew point and delay, and not causal statements.

HW-1.Rmd
```diff
@@ -65,48 +145,331 @@ Please indicate
...
+filter(!is.na(year)) %>%
+group_by(carrier) 
 
+#Plot 02
+g <- ggplot(data = total_planes, aes(x= "", y= year)) 
+g + 
+  ggtitle("Plot #02 \n Age Distribution of planes by airline") +
+  xlab("Carrier") + ylab("Year of Manufacture") + 
+  geom_boxplot() +  geom_hline(yintercept=1980, col="red", size=0.5) +
+  facet_grid(.~carrier)
```
rudeboybert: Next time consider sorting the boxplots by median age. I'll show next wednesday how to do this.

HW-1.Rmd
```diff
@@ -65,48 +145,331 @@ Please indicate
...
+  filter(!is.na(avg_delay)) %>% 
+  filter(!is.na(conditions))
+ 
+#Plot 07
+m <- ggplot(data = delay_conditions, aes(x= conditions, y = avg_delay))
+m + 
+  xlab("Condition") + 
+  geom_boxplot() + 
+  coord_flip() +
+  scale_y_log10() +
```
rudeboybert: You have to watch out what log10'ing the data here does. Because log is undefined for values less than or equal to 0, R will automatically throw these out. So this changes the population for which you're making statements about. It goes from:
* all flights
* to only those flights with a delay > 0

A safer approach is to explicitly remove these from the `delay_conditions`, make a comment, and then mention this in your write-up.

HW-1.Rmd
```diff
@@ -65,48 +145,331 @@ Please indicate
...
+a box-and-whisker plot (Plot #08) was created to demonstrate the effects of each
+condition on the time of delay. The x-axis shows the log10 of the delay time 
+(min) in order to ease viewing of the plots. At first glance, one can see that 
+the medians from light freezing rain and freezing rain resulted in the longest 
+flight log10-delays (min). Freezing rain can be quite problematic for a flight- 
+especially for one in Houston, where freezing conditions are not common. 
+Freezing conditions can cause problems with the runway, cleaning and water 
+loading, and engines. The plane actually has to be sprayed to prevent freezing 
+while the plane is in flight. In addition, in-flight conditions could be 
+effected, so waiting for the freezing rain to perhaps subside is a reason for
```
rudeboybert: How many days of freezing rain does Houston get?

HW-1.Rmd
```diff
@@ -65,48 +145,331 @@ Please indicate
...
 
-
 ## Question 5:
 
 I want to know what proportionately regions (NE, south, west, midwest) each 
 carrier flies to from Houston in the month of July.  Consider the `month()`
 function from the `lubridate` package.
 
+''
+From the data given, the figure (Plot #08) from the code below was made to illustrate the proportion of flights to four geographical regions for each carrier in the month of July. Airlines YV, FL, and AA fly only to states in the SOUTH, while the airline B6 flies only to the NORTHEAST, AS and F9 fly only to the WEST. B6 is Jetblue, who has a hub in New York City- they only fly to New York from Houston. AS is Alaskan Airlines, who only flies to Seattle from Houston, and F9 is Frontier Airlines which flies to the Denver from Houston. The other airlines are a mixture of midwest, NE, south, and West. One can see trends. For example, the most flied to region is the South. The least flied region is the NE. The west is the second most flied to region, and the midwest is the third most. Geographically, this make sense. The South is the closest region to Texas, then the West, then midwest, and then the NE. The Northeast is the farthest region away from Texas.
```
rudeboybert: when talking about repeated cases of specifics, I find it helps to use bullet points, instead of large paragraphs. The point you bring up are great! So something like

* B6 is Jetblue, who has a hub in New York City- they only fly to New York from Houston. 
* AS is Alaskan Airlines, who only flies to Seattle from Houston
*  etc

HW-1.Rmd
```diff
@@ -65,48 +145,331 @@ Please indicate
...
+  filter(!is.na(conditions))
+ 
+#Plot 07
+m <- ggplot(data = delay_conditions, aes(x= conditions, y = avg_delay))
+m + 
+  xlab("Condition") + 
+  geom_boxplot() + 
+  coord_flip() +
+  scale_y_log10() +
+  ylab("Delay log10 (min)") +
```
rudeboybert: nit-pick. its not quite `"Delay log10 (min)"` but `"Delay in min (log10-scale)"`

rudeboybert:
**OVERALL COMMENTS**: things look good. The one comment is sometimes your interpretations involved describing things one can already plainly see. One needs to be judicious and only verbally pick out salient highlights. I truly believe in the "less is more" principle here.

Coding wise, things could be tightened up a bit. It got a bit hard to read at times, but nothing disastrous:

* Not having fragmented pieces of code working on the same data frame
* Clean up blank spaces

My feedback is best viewed in [Split Screen Mode](https://github.com/Middlebury-Data-Science/HW-1/pull/14/files)

PR #17, opened 2016-03-10
Sanket Vadlamani

uvadlamani:
Homework 1


HW-1.Rmd
```diff
@@ -53,12 +55,34 @@ Please indicate
 * Which seasons did we tend to see the most and least delays of > 30 minutes.
 
 ```{r, echo=FALSE, message=FALSE, fig.width=12, fig.height=6}
+delay_or_not <- c(30:982)
+flights <- flights %>%
+  mutate(delay = ifelse(dep_delay %in% delay_or_not, 'yes', 'no'))
```
rudeboybert: you could've just done `delay > 30` instead of using an `%in%`

HW-1.Rmd
```diff
@@ -53,12 +55,34 @@ Please indicate
...
 
+proportion_flights_delayed <- left_join(delayed_flights_per_month, total_flights_per_month, by="month") %>% 
+mutate(prop = round(delayed_flights_per_month/total_flights_per_month,3))
 
+proportion_flights_delayed <- filter(proportion_flights_delayed, delay == "yes")
 
+ggplot(data=proportion_flights_delayed, aes(x = month, y = prop)) + geom_bar(stat="identity")
+```
+
+The plot above is a time series of the proportion of flights that were delayed on average per month. It is evident that the months of April, May, June and July had the most delays. Therefore summer on average had the most delays. It also seems like fall had the least delays.Months August, September, October and Novemeber had the least delays. The limitation of this plot is that it bins the delays by month and therefore is not as accurate as a similar graph with date as the unit of the time axis. However having 365 values presents difficulties in interpretation. Since we are only concerned with season it is justifable to use months instead, even though we cannot convey the same amount of information, because most people would think of seasons in terms of months and not dates themselves.
```
rudeboybert: ah ok, you talk about it here. While there is no right answer, I would argue that showing a time series day by day, along with a smoother, does the job. We'll talk about this on Wednesday.

HW-1.Rmd
```diff
@@ -66,14 +90,18 @@ Some people prefer flying on older planes.  Even though they aren't as nice,
...
-
+planes_and_flights <- inner_join(planes, flights, by="plane")
+planes_and_flights_selected <- planes_and_flights %>%
+  arrange(year.x, carrier) %>% 
+  select(year.x , carrier)
+planes_and_flights_selected_unique <- unique(planes_and_flights_selected)
+
+ggplot(data=planes_and_flights_selected_unique, aes(x = carrier, y = year.x)) +
+  geom_hline(yintercept=1980) +
+  geom_point()
```
rudeboybert: your thinking is good, in that you are not just presenting the average age, but showing the spread of the different ages. The problem with using `geom_point` is that we don't know how points are superimposed on top of each other. i.e. how many Delta (DL) planes are from 1980? One, two, or ten?  Either you should
* use `geom_jitter()` so that we add a little random noise to the points so that we know how many points are involved
* better yet, do boxplots! they give summary measures for a set of points (1 quart, median, 3rd quart) but use much less ink.

HW-1.Rmd
```diff
@@ -66,14 +90,18 @@ Some people prefer flying on older planes.  Even though they aren't as nice,
...
 ```
 
-
-
-
-
-
+In the graph above each point represents an aircraft. The x axis represents the carrier that owns that particular aircraft. The y axis represents the year in which that aircraft was built. A quick online search suggests that it was from that 1980 most planes started to decrease their leg-room. Therefore 1980 can be the parameter used to distinguish easily between "old" and "new" flights. Passengers who prefer flying older airplanes for the extra leg-room should choose American Airlines as AA is the carrier that has most number of flights in operation that were built before 1980. The other option is Delta though not as much as American.
```
rudeboybert: Good value adding with the 1980 quote.

HW-1.Rmd
```diff
@@ -82,24 +110,87 @@ they tend to have more room.  Which airlines should these people favor?
...
+  distinct(flight)
+
+south_west_flights_distinct_destination <- south_west_flights_distinct %>% 
+  group_by(dest)  %>% 
+  tally() %>% 
+  rename(south_west_flights_distinct_destination = n)
+
+airports_and_states <- left_join(airports,states,by=c("state"="state"))
 
+south_west_flights_destination_state <- left_join(south_west_flights_destination, airports_and_states, by=c("dest"="iata")) %>%
```
rudeboybert: It's considered good coding practice to keep lines under 80 characters, so that they don't bleed off the page, especially for laptops. For example, I would code line 131 as:

```r
south_west_flights_destination_state <- 
  south_west_flights_destination
  left_join(airports_and_states, by=c("dest"="iata")) %>% 
  select(dest,south_west_flights_destination,state,fullname)
```

HW-1.Rmd
```diff
@@ -82,24 +110,87 @@ they tend to have more room.  Which airlines should these people favor?
...
+
+airports_and_states <- left_join(airports,states,by=c("state"="state"))
 
+south_west_flights_destination_state <- left_join(south_west_flights_destination, airports_and_states, by=c("dest"="iata")) %>% 
+  select(dest,south_west_flights_destination,state,fullname)
+south_west_flights_distinct_destination_state <- left_join(south_west_flights_distinct_destination, airports_and_states,by=c("dest"="iata")) %>% 
+  select(dest,south_west_flights_distinct_destination,state,fullname)
 
+ggplot(data=south_west_flights_destination_state, aes(x = state, y = south_west_flights_destination)) + geom_bar(stat="identity")
+ggplot(data=south_west_flights_distinct_destination_state, aes(x = state, y = south_west_flights_distinct_destination)) + geom_bar(stat="identity")
```
rudeboybert: You definitely want to reorder the bars in order of frequency, and fall victim to the ALABAMA FIRST trap.

HW-1.Rmd
```diff
@@ -82,24 +110,87 @@ they tend to have more room.  Which airlines should these people favor?
...
+  group_by(conditions) %>% 
+  tally() %>% 
+  rename(delays_by_condition = n) %>% 
+  arrange(desc(delays_by_condition))
+delays_by_condition_2
+
+ggplot(data=delays_by_condition_2, aes(x = conditions, y = delays_by_condition)) +
+  geom_bar(stat="identity")+
+  coord_flip()+
+  ggtitle("Figure 4")
```
rudeboybert: * Use more informative titles. Readers should have to spend the least amount of time possible to know what you're showing
* Same point as earlier about using `geom_point` instead of something like `geom_boxplot`
* Also, raw R tables are jarring to read. Consider using the `kable()` function in the `knitr` package to convert R data frames to markdown tables, to make the output look nice.

HW-1.Rmd
```diff
@@ -108,5 +199,51 @@ carrier flies to from Houston in the month of July.  Consider the `month()`
...
+  filter(month == 7, region=="west") %>%
+  group_by(carrier) %>% 
+  tally() %>% 
+  rename(total_flights_in_July_west=n)
+
+july_flights_midwest <- airports_and_states_and_flights %>% 
+  filter(month == 7, region=="midwest") %>%
+  group_by(carrier) %>% 
+  tally() %>% 
+  rename(total_flights_in_July_midwest=n)
```
rudeboybert: Here, you didn't need to make a separate table for each region.  You can group by more than one variable, i.e.
```r
group_by(carrier, region)
```

HW-1.Rmd
```diff
@@ -108,5 +199,51 @@ carrier flies to from Houston in the month of July.  Consider the `month()`
...
+total_4<- full_join(total_3,july_flights,by="carrier")
+
+
+proportion <- total_4 %>% 
+  mutate(prop_NE = round(total_flights_in_July_NE/total_flights_in_July,3)) %>%
+  mutate(prop_south = round(total_flights_in_July_south/total_flights_in_July,3)) %>% 
+  mutate(prop_west = round(total_flights_in_July_west/total_flights_in_July,3)) %>% 
+  mutate(prop_midwest = round(total_flights_in_July_midwest/total_flights_in_July,3)) %>%
+  select(carrier,prop_NE,prop_south,prop_west,prop_midwest)
+proportion
```
rudeboybert: While this table is not too big so its digestible, really a visualization gets the point across so much faster.  Also, see my note earlier about making tables look nice in Markdown using the `kable()` function from the `knitr` package or the DT package https://rstudio.github.io/DT/

rudeboybert:
**OVERALL COMMENTS** Your visuals could use a bit more polish, but good first effort.

Coding style: your code was a bit hard to read. Try to be consistent with
* line breaks
* keep lines under 80 characters
* using piping and only creating the minimal number of new data frames needed to get the job done.

My feedback is best viewed in [Split screen mode](https://github.com/Middlebury-Data-Science/HW-1/pull/17/files)

PR #20, opened 2016-03-14
Shaojin Li

ShaojinLi:
HW1

### DIFF
--- a/HW-1.Rmd
+++ b/HW-1.Rmd
@@ -1,47 +1,42 @@
 ---
 title: "MATH 216 Homework 1"
-author: "WRITE YOUR NAME HERE"
+author: "Shaojin Li"
 output: html_document
 ---
 
-```{r, echo=FALSE, message=FALSE, cache=TRUE}
+```{r, echo=FALSE, message=FALSE}
 # DO NOT EDIT THIS SECTION
-
 # Load packages
 suppressPackageStartupMessages(library(ggplot2))
 suppressPackageStartupMessages(library(dplyr))
 suppressPackageStartupMessages(library(lubridate))
-
+```
+```{r, echo=FALSE, message=FALSE, cache=TRUE}
 # Load data sets. Note this assumes this .Rmd files is in the same directory as
 # the .csv files.
 flights <- read.csv("data/flights.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df() %>%
-  mutate(date=as.Date(date))
+ tbl_df() %>%
+ mutate(date=as.Date(date))
 weather <- read.csv("data/weather.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df() %>%
-  mutate(date=as.Date(date))
+ tbl_df() %>%
+ mutate(date=as.Date(date))
 planes <- read.csv("data/planes.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df()
+ tbl_df()
 airports <- read.csv("data/airports.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df()
+ tbl_df()
 states <- read.csv("data/states.csv", stringsAsFactors = FALSE) %>% 
-  tbl_df()
+ tbl_df()
 ```
-
-
-
 
 
 ## Admistrative:
 
 Please indicate
 
-* Who you collaborated with:
-* Roughly how much time you spent on this HW:
-* What gave you the most trouble:
+* Who you collaborated with: I did it by myself
+* Roughly how much time you spent on this HW: 15 hrs
+* What gave you the most trouble: Plot data in better ways
 * Any comments you have:
-
-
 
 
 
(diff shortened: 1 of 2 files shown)


HW-1.Rmd
```diff
@@ -53,12 +48,60 @@ Please indicate
...
+  full_join (flight_delayed_month, flight_total_month, by="month") %>%
+  mutate (prop_delayed_month=num_flight_delayed/num_flight_total)
+
+Q1_month<-ggplot(flight_delayed_month, aes(x=month, y=prop_delayed_month))+
+  geom_line ()+
+  ggtitle ("Delayed Flights in 2011 by Month")+
+  xlab ("Month")+
+  ylab ("Proportion of Flights Delayed by 30 Mins or More")+
+  scale_x_continuous(breaks=c(1:12))
+Q1_month
```
rudeboybert: I would argue that including the `geom_smooth` in the first plot makes the 2nd plot somewhat redundant. I like the "less is more" approach to presenting results; if its redundant, leave it out.

HW-1.Rmd
```diff
@@ -53,12 +48,60 @@ Please indicate
...
 
 ```
-
-
-
-
+By combining graph "Q1_day" and graph "Q1_month", We can see the most delays of >30 minutes in the late Spring and early Summer months, especially June and July. The least dalays of >30 minutes occur in Autumn. Summer months also tend to have more long delays than the winter months.
```
rudeboybert: I wouldn't refer to `Q1_day` and `Q1_month` in your write up. Someone just looking at the output is not going to what these refer to.

HW-1.Rmd
```diff
@@ -67,12 +110,29 @@ they tend to have more room.  Which airlines should these people favor?
...
+Q2<-ggplot(flights_year, aes(x=carrier, y=year))+
+  geom_point(size=1)+
+  xlab("Carrier")+ 
+  ylab("Year")+
+  ggtitle("Age of Planes by Airline")
+Q2
 
 
+```
+From the table "flights_year_stats" we can see that on average, airline MQ has the oldest planes. However, from the graph "Q2" we can also see that airline MQ only has two planes. By combinding both the graph and the table, we can see that AA has the most old planes, followed by Delta. AA has less new planes and smaller standard deviation of the year. DL has more planes, which means that DL might either have more older planes or more newer planes. A much better way to find which airlines have more older planes is to "weight" the year by the number of different planes that each airline has for each year, but I did not know how to do this.
```
rudeboybert: I would've have used `geom_point` here, b/c as you say you need to know the number of planes for a given airline for a given year, but you can't tell this b/c you don't know how many points are superimposed on top of each other.

A better tool would've been `geom_boxplot`. That way you can see what the 1st quartile/median/3rd quartile are.

HW-1.Rmd
```diff
@@ -80,14 +140,39 @@ they tend to have more room.  Which airlines should these people favor?
...
+  ylab("Number of Listed WN Flights")+
+  ggtitle("States that Listed Southwest Airlines Flights Fly to")
+Q3_listed
+  
+Q3_all<-ggplot(WN_all, aes(x=state, y=WN_all))+
+  geom_bar(stat="identity")+
+  xlab("State")+ 
+  ylab("Number of All WN Flights")+
+  ggtitle("States that All Southwest Airlines Flights Fly to")
+Q3_all
```
rudeboybert: Your plots all suffer from the ALABAMA FIRST issue, i.e. alphabetical order is not the best way to order the bars.  Try out the `reorder` function like I do here:

```r
Q3_all<-ggplot(WN_all, aes(x=reorder(state, WN_all), y=WN_all))+
  geom_bar(stat="identity")+
  xlab("State")+ 
  ylab("Number of All WN Flights")+
  ggtitle("States that All Southwest Airlines Flights Fly to")
Q3_all
```

HW-1.Rmd
```diff
@@ -80,14 +140,39 @@ they tend to have more room.  Which airlines should these people favor?
...
 
+```
+Listed Southwest Airlines flights tend to fly mostly to Texas, Florida, Louisiana, Oklahoma, and Illinois. The difference between each states is not very distinct. All Southwest Airlines flights tend to fly to mostly Texas, followed by Florida and Louisiana. For all Southwest Airlines flights, the number of flights to Texas is significantly larger than the number of flights to other states.
```
rudeboybert: Yes, mainly in the south.

HW-1.Rmd
```diff
@@ -95,11 +180,24 @@ What weather patterns are associated with the biggest departure delays?
...
+  arrange(desc(avg_delay)) %>%
+  na.omit() 
+
+Q4<-ggplot(flights_weather, aes(x=conditions, y=avg_delay))+
+  geom_bar(stat="identity")+
+  theme(axis.text.x=element_text(angle=30, hjust=1))+
+  xlab("Weather")+
+  ylab("Average Flight Delay (min)")+
+  ggtitle("Flight Delays by Weather")
+Q4
```
rudeboybert: Great job tilting the labels so that they are legible! 

But let me ask you: the bars for "Freezing Rain" and "Light Freezing Rain" are the biggest, but how often is there Freezing Rain in Houston?  Is this something they need to worry about 6 months of the  year?  One or two days a year?  In terms of informing decisions, some notion of how often these patterns occur matters too.

HW-1.Rmd
```diff
@@ -109,4 +207,40 @@ function from the `lubridate` package.
...
+carrier_region<-mutate(carrier_region, prop=(flights_by_carrier_region/flights_by_carrier))%>%
+  arrange(desc(prop))
+
+Q5<-ggplot(carrier_region, aes(x=carrier, y=prop, fill=region))+
+  geom_bar(stat="identity")+
+  xlab("Carrier") + 
+  ylab("Proportion of Flights by Region") +
+  ggtitle("Proportion of Flights by Region in July")
+Q5
+
```
rudeboybert: I see you used the `na.omit()` command. Awesome, I didn't know about this command!

However, this leads to gaps in the bars. Like why is FL's and CO's bars shorter than the others? You should either have a 5th category for missing, or describe in your write-up why these are missing.

Great observation connecting the # of destinations to the # of flights! Way to connect insight from another question to this one. I highly suggest you volunteer this during tomorrow's session.

HW-1.Rmd
rudeboybert: HAHA, I just saw your comment in your code about the 2nd graph. Again, this is a perfect example how in data analysis, there are no "right" ways to do things. You just got to pick your way, and be confident about it!

rudeboybert:
**Overall comments**: great job. A few improvements in the visualizations can be made here and there, but a great first assignment.

Also, your code is super legible and easy to read and well indented.  Maybe throw in a few small comments for each piping chain explaining what the resulting data frame is.

My feedback is best viewed in [Split Screen Mode](https://github.com/Middlebury-Data-Science/HW-1/pull/20/files)

